PR #25, opened 2026-08-17
feat(ctx): vendor usage monitoring, use_credits gating, pace-to-reset throttle

Glubiz:
## Summary

zirv now monitors vendor usage for both harnesses and paces itself against the subscription window:

- **Usage in the header** — wrap's status bar and the dashboard header show per-harness usage (codex via a 60s-throttled, filesystem-only rollout scan; never HTTP on a redraw path).
- **Hybrid data source** — passive collectors are primary (claude statusline tee; codex session-rollout JSONL scan in `window.rs`), with an active HTTP poll fallback (`poll.rs`, `ureq`/rustls) consulted only when the stored reading is stale, floored to one attempt per `poll_min_interval_secs` via marker files. The Anthropic OAuth usage endpoint is fixture-verified; the codex ChatGPT endpoint ships best-effort unverified (recorded in Known Issues).
- **Pace-to-reset throttle** — `PaceDecision::Slow` spreads the remaining budget linearly over time-to-reset (`D = t_rem*(p-soft)/(max-soft)`) between `soft_percent` and `max_percent`, with the existing hard pause at max. The slow deadline is monotonic across re-checks (test-pinned), so a throttle can never silently stretch into a full park.
- **`use_credits` gate** — per-harness, default **disabled** for both. Enabled, it skips zirv's proactive throttle/pause only; a vendor-reported limit hit still parks. `pace.use_credits` and the poll keys are `REPO_FORBIDDEN`: a repo checkout cannot grant itself a credits exemption or change polling. `pace.enabled = false` is a master off-switch for all proactive vendor requests.
- **Session conventions v2** (`prompt.rs`) — two new injected rules (verify once then trust the result; keep scope to what was asked, prefer simple solutions), delivered to claude sessions via the composed prompt and to codex workers via a new task-prompt fallback.

Token hygiene: OAuth tokens are read fresh per call and exist only in the one Authorization header — never logged, persisted, or derivable; tests are hermetic (stub pollers, redirected home dirs; two live-HTTP-in-tests holes were caught and closed in review).

Version 2.9.0.

## Review

Per-task review gates (9 tasks), a whole-branch review (all nine integration invariants hold; requirements traced end-to-end), and a native /code-review pass whose one completed verifier's finding (usage verb polled under `ZIRV_CTX_PACE=false`) is fixed here.

**Residual findings** (recorded, not blocking):
- A failed poll attempt degrades silently where the spec promised a one-time announcement — recorded deviation in Known Issues; proper fix wants a `maybe_poll` return-type change + mockable transport.
- The /code-review pass was cut short by a session usage limit after confirming one finding; 7 deduped candidates went unverified and can be re-reviewed post-merge.
- The codex cross-harness review seat could not run (codex CLI fails on a local auth/JSON parse error before any work).
- Deferred cleanups: stale `#[allow(dead_code)]` markers in poll.rs/window.rs, a scan floor for long parks, the fresh-claude statusline hint no longer surfacing.


Glubiz:
**Post-open review round 2** (rerun of /code-review high after the usage-limit reset, per request): 8 finder angles → 48 raw candidates → 20 adversarially confirmed. The 10 reported correctness findings are addressed in three commits now pushed:

- `82fa28b` — pace gate: estimator-only pacing is no longer bypassed by the no-source early-return (which also un-wedges the limit-park hot-relaunch loop for API-key/Bedrock setups); a latched Slow deadline survives the reading going stale mid-wait (the old test was passing *through* that escape and now genuinely pins the monotonic rule); a fresh reading whose `resets_at` already passed no longer produces a phantom post-rollover throttle; Slow episodes announce once on the `zirv ▸` channel; park-loop rollout scans are floored to 60s.
- `b8b50a6` — rollout timestamps bounded to 4-digit years (closes a debug-build overflow panic reachable from wrap's status-bar redraw); `refresh_codex_usage` skips rewriting identical state; the ureq agent is built once with a 3s connect timeout distinct from the 10s overall; fresh claude machines get the actionable `zirv ctx usage tee` guidance back instead of the generic no-source line.
- `d4c85b8` — CLAUDE.md no longer claims `soft_percent` is repo-forbidden (the spec rules it a tuning knob; a repo can already set `pace.enabled=false`); Known Issues records that polling is structurally inert on keychain/API-key setups and names the tee as the remedy.

**Verified-but-deferred cleanups** (cut by the report cap, all quality not correctness): ~27 stale `#[allow(dead_code)]` attributes in poll.rs/window.rs/state.rs, the four-way `~/.codex/sessions` path resolution duplication, six copy-pasted conventions-fallback gates, four hand-assembled PaceGate sites, duplicate timestamp parsers, `usage.rs` re-implementing `pace::refresh_sources`, ureq's unneeded default gzip feature, a benign cross-process poll-floor TOCTOU, and conventions-vs-handoff-note ordering on nudge/rot relaunches. This round was the second and final fix round; these are recorded residuals.

PR #29, opened 2026-08-19
feat(dash): adaptive input poll and a Ctrl+A ? help overlay

### LINKED ISSUE #30: zirv ctx inbox: mail silently consumed without being surfaced to the polling caller
## Summary

While polling `zirv ctx inbox` from a script for a reply from a `zirv agent codex` dashboard-pane worker, the reply mail ended up moved to the `read/` archive (i.e. something consumed it) without ever being printed by any of my own `zirv ctx inbox` calls — the polling loop reported a timeout, and two subsequent manual checks (`zirv ctx inbox` and `zirv ctx inbox --peek`) both returned nothing, even though the message was later found sitting in the `read/` folder on disk.

## Timeline (repo: `D--GitHub-zirv-fitness-tracking`, my session `9690373b-5f3e-4749-97b2-37ce036dd929`)

1. `zirv agent codex "<review prompt>" -- --model gpt-5.6-terra` was run from inside a dashboard session, which spawned an attached pane (`8f44f4db`) rather than running headless-to-completion, per the documented dashboard behavior.
2. `zirv ctx status` showed a decision-log entry for that pane: `{"action":"prompt-skipped","detail":"agent has no verified system-prompt mechanism (unsupported)"}` — flagged as a "degraded surface: codex (launch-level: no rot score, usage, turn signal, or injected prompt)" in `zirv ctx --help`. My working theory is this means the "report back by mail when done" instruction that's normally injected for background workers never reached the codex pane, since I never wrote such an instruction myself.
3. I ran a background poll loop (~10 min, 15s interval) calling `zirv ctx inbox` in a `while` loop, checking for non-empty output not matching `no mail|0 unread|^$`. It timed out with no message ever printed.
4. I then ran `zirv ctx nudge 8f44f4db --message "..."` asking the pane to send its findings via `zirv ctx send`. Nudge sent at unix ts `1787147748` (confirmed via the mail file below).
5. I ran a second background poll loop (5 min, 15s interval), same logic as step 3. It **also timed out** with "no mail received" printed as the final line after all 20 iterations completed.
6. The user (watching the dashboard directly) then told me "Codex has responded."
7. I ran `zirv ctx inbox` manually — empty output. I ran `zirv ctx inbox --peek` — also empty output.
8. I searched the state dir directly (`C:\Users\josj\AppData\Local\zirv\ctx\mail`) and found the reply **already sitting in `read/`**:
   ```
   C:/Users/josj/AppData/Local/zirv/ctx/mail/D--GitHub-zirv-fitness-tracking/read/1787147843-unknown.md
   ```
   Sent ts `1787147843` — only 95 seconds after the nudge (ts `1787147748`), which is well inside the second poll loop's 300-second window and should have been picked up by at least one of its ~20 fifteen-second-interval checks.

## Impact

A message that should have been retrievable via the documented `zirv ctx inbox` CLI was moved to the read archive without ever being returned to any caller I made (looped or manual), effectively losing it from the perspective of the intended recipient session. I only recovered the content by reading the raw `.md` file out of the state directory directly, which isn't a documented/supported interface.

## Suspected contributing factors (unconfirmed — flagging for investigation, not asserting root cause)

- Possibly a race between my `run_in_background` Bash-tool subprocess polling loop and some other process/hook belonging to the same session (or the interactive `zirv ctx chat` orchestrator itself, pid 342916 for session `9690373b`) that also reads/consumes `zirv ctx inbox` as part of its own mail-surfacing mechanism, such that the "other" consumer wins the race and the content is dropped rather than being handed to whichever caller actually needed it.
- Possibly related to the "degraded surface: codex ... no verified system-prompt mechanism" limitation noted above — if codex-origin mail is tagged/routed differently (note the `From-session: unknown` field on the codex reply, vs. a real UUID for claude-origin mail), that could interact badly with whatever matches mail to a waiting reader.

## Environment

- `zirv version`: 2.12.0
- OS: Windows 11 (via Git Bash)
- Reproduced via Claude Code session orchestrating a `zirv agent codex` dashboard-pane worker

## Suggested fix direction

At minimum, `zirv ctx inbox` (and whatever internal process may also be consuming mail concurrently) should not be able to silently drop a message between "arrived" and "any reader observes it." A `--peek`-first read before any consuming read, and/or a single consistent consumer per session (rather than multiple independent readers racing on the same mailbox), would prevent this class of loss. If session-side automatic mail injection into agent context exists, it should log when it consumes something, so at minimum the loss is diagnosable from the decision log rather than silent.

Glubiz:
## What

- **Instant-feeling input**: the dashboard event loop's flat 50ms `event::poll` wait is now adaptive — 10ms inside a 300ms hot window opened by operator key/mouse input, 50ms when idle. Keystroke echoes repaint within ~10ms instead of up to 50ms. Pane output never refreshes the window, so a streaming pane can't hold the loop at the fast rate; idle CPU is unchanged.
- **Help overlay**: `Ctrl+A ?` (or `h`) shows the full keymap, any key closes it. Content is a compile-time table with tests pinning every row to real `filter_key` dispatch, exhaustive `DashAction` coverage (a new action can't ship without a help-row decision), and an 80x24 fit budget.
- **Mouse drag text selection**: left click-drag over a pane (whose child hasn't claimed mouse reporting) now selects text — `?1002` drag tracking, REVERSED highlight, and copy-to-clipboard via OSC 52 on release (64 KiB cap, best-effort). A selection is cancelled by scrollback movement, any resize/zoom, or new output on the selected pane, so stale coordinates can never copy the wrong text. Click-without-drag copies nothing; mouse-mode children keep the old forward-to-child behavior (native shift+drag still works there).
- **Shift+Enter newline fix**: `encode_key` treats Enter+ALT like Enter+SHIFT and emits `ESC CR`. An empirical ConPTY probe of the real claude CLI established that `\x1b\r` is the one encoding claude natively treats as newline, and that Windows Terminal's `/terminal-setup` rewrite of Shift+Enter arrives at zirv as a folded Enter+ALT event — which previously degraded to a bare `\r` and submitted the prompt instead of inserting a newline.
- **Worker mail can no longer be silently lost** (fixes #30): spawned worker panes always get a session identity in their child env (codex panes previously had none, so replies went out as `From-session: unknown` broadcasts that any sweep could claim into the wrong pane and archive unseen). Report-back mail is therefore directed (`--to-session`), visible only to its addressee, and every on-behalf-of mail consumption (exec/loop drains, dash sweep, spawn harvest, mail overlay) now writes a `mail-consumed` decision-log entry so any future loss is diagnosable.

## Why

Typing in a `zirv chat` pane felt debounced; text couldn't be selected or copied out of a pane at all; Shift+Enter submitted instead of newlining; and a codex worker's review reply vanished into the mail read-archive without ever being surfaced (#30).

Includes the Obsidian vault updates for all changes and a version bump to 2.13.0. Known accepted edges: terminals shorter than ~22 rows clip the help overlay's tail; terminals without OSC 52 support silently skip the clipboard copy (documented in Known Issues).


Glubiz:
## Follow-up: mouse text selection in dashboard panes (decided, not yet built)

**Problem (root cause confirmed):** the dashboard enables terminal mouse reporting (`?1000h?1006h`, written in `dash/mod.rs` gated by `dash.mouse`, default on) for wheel scroll and click forwarding. The terminal then routes all mouse events to zirv, so native click+drag selection does nothing in any dash pane — and Apple Terminal offers no bypass modifier at all. Drag events are never even generated, since dash deliberately never enables motion tracking (Decision Log 2026-08-15). The wrap/chat path is byte-transparent and is NOT the cause (bare `claude` selection works fine).

**Decision (Jonathan, 2026-08-19):** build true in-pane selection in dash, rather than a mouse-off toggle or flipping the `dash.mouse` default. Sketch for whoever picks this up:

- Enable `?1002h` (button-drag tracking) alongside `?1000h?1006h`. This partially revisits the 2026-08-15 decision-log entry: that entry rejected `?1003h` all-motion flood; `?1002h` only reports while a button is held, so the flood argument doesn't apply. Update the decision log accordingly.
- Track Down -> Drag -> Up in pane-local coordinates (`pane_local_mouse` in `dash/mod.rs` already maps Down/Up to a pane).
- Render the selection highlight over the vt100 cells (`render_grid` in `dash/ui.rs` already has per-cell access).
- On release, extract the selected text from the pane's vt100 screen and copy it: `pbcopy` on macOS, OSC 52 as the portable path (no clipboard code exists anywhere in the repo yet — greenfield).
- Define behavior when the pane is in the wheel-scrolled scrollback view vs the live screen.
- `dash.mouse = false` must keep skipping all of it, and the repo rule stands: never `crossterm::EnableMouseCapture`; raw escape writes via `term.rs` only (see Known Issues).
- Add the resulting gesture to the `Ctrl+A ?` help overlay introduced in this PR.

This PR intentionally ships without it; the note exists so the follow-up lands as its own reviewed change.

Issue #115, opened 2026-08-24
Dash-spawned codex workers never send their report-back mail on completion


Glubiz:
## Symptom

A codex worker spawned from a dashboard session with `zirv agent <name> "<prompt>"` is supposed to report its outcome back to the spawning session by mail when it finishes (`zirv ctx inbox`). For codex workers this report-back reliably fails to arrive: the worker visibly completes its task in its pane, but the requesting session's inbox stays empty (`zirv ctx status` → `mail: 0 unread`) and no `[zirv ▸ mail]` wake-up ever fires. The operator reports this as a recurring problem across sessions.

## Observed instance (2026-08-24, zirv 2.25.x, Windows 11)

1. Claude session `0f1db298` (dash) ran `zirv agent codex "<code-review prompt>" -- --model gpt-5.6-terra` → pane `f7988311`.
2. The worker completed the review in its pane. No mail arrived; inbox empty ~10+ minutes after completion, pane still live.
3. Only after `zirv ctx nudge f7988311 --message "send your verdict via zirv ctx send --to-session 0f1db298 ..."` **plus a manual Enter in the pane** (see #114) did the worker run `zirv ctx send` and the mail arrive correctly.

Step 3 shows the mail transport itself works and the worker is willing and able to send — the failure is that the report-back contract is never actually triggered on completion.

## Likely causes (both worth checking)

1. **Delivery of the instruction** — if any part of the spawn-time "report back by mail when you finish" contract is delivered into the pane via the same input-injection path as `ctx nudge`, #114's missing-Enter bug would silently drop it for codex panes.
2. **Contract enforcement** — even with delivery fixed, nothing verifies the worker actually sent its completion mail. Options: the dash supervisor detects worker turn-completion and sends the mail on the worker's behalf (harness-side, deterministic — the supervisor already scores the transcript, so it knows when the worker finished and could extract/forward the final message), or it detects "finished but no mail sent" and re-prompts.

A harness-side completion report (option 2, supervisor-forwarded) would make the contract deterministic instead of depending on an agent remembering an instruction — worth considering as the real fix, with the prompt-layer instruction kept as the content source.

## Impact

Cross-harness delegation (`zirv agent codex ...`) from a dashboard is effectively fire-and-forget-and-poll: the spawning agent waits for mail that never comes, and a human has to notice and intervene. This breaks the documented workflow (worker "is instructed to report its outcome back to this session by mail when it finishes").

## Verification requirements for a fix

[...]

PR #116, opened 2026-08-24
fix(dash): submit pane injections reliably and remind workers to report back (#114, #115)

### LINKED ISSUE #114: zirv ctx nudge into a codex dash pane types the message but never presses Enter
## Symptom

`zirv ctx nudge <pane> --message "..."` targeting a **codex** dash pane types the message into the codex TUI's input field but never submits it: Enter is not pressed (or the CR is swallowed by the TUI), so the message sits unsent in the input box until a human presses Enter in that pane. The operator has observed this repeatedly; it silently breaks every workflow that depends on a message actually reaching a codex worker.

## Reproduction (observed 2026-08-24, zirv 2.25.x, Windows 11, dashboard session)

1. From a claude chat session running under a dashboard, spawn a codex worker: `zirv agent codex "<review prompt>" -- --model gpt-5.6-terra` → `spawned in dashboard as f7988311`.
2. Later, nudge it: `zirv ctx nudge f7988311 --message "..."` → CLI reports `queued for f7988311 (codex, dash)`.
3. Observe the pane: the message text appears in the codex input field, but is never submitted. The worker never sees it.
4. Manually press Enter in the pane → the worker immediately processes the message (in this case it then ran `zirv ctx send --to-session 0f1db298 ...` and the mail arrived as instructed).

Step 4 is the discriminator: the text delivery works, only the submit keypress is missing.

## Expected

A nudge delivered to a live codex pane reaches the agent without human interaction — the injection path must send the submitting keypress, and send it in a way the codex TUI actually registers (codex's TUI is known to swallow a CR that arrives in the same write as pasted text; the submit likely needs to be a separate, slightly delayed write, similar to bracketed-paste handling).

## Suspected area

The dash pane input-injection path (`src/commands/ctx/dash/pane.rs` / the nudge delivery in `src/commands/ctx/`), specifically the codex adapter's submit sequence. Claude panes may or may not share the flaw — needs checking; only codex panes were observed.

## Verification requirements for a fix

This is harness-facing (`ctx nudge`/dash/pane), so per repo convention the fix needs the Docker AI-feature command matrix run (ctx chat/agent/dash/mail/hook) before the PR, plus the Linux PTY test pass (`wrap::` family, non-root, `git -c core.autocrlf=false archive` export).

## Consequence

Downstream, this is the likely root cause of dash-spawned codex workers "never mailing their report back" — tracked separately (see linked issue).

Glubiz:
Fixes #114 and #115.

**#114 — pane injections never submitted in codex panes.** `zirv ctx nudge`/mail/report injections wrote the visible line and its trailing CR in one pty write; codex's TUI paste-burst heuristic folded the CR into the pasted text as a newline, so messages sat unsubmitted until a human pressed Enter. `Pane::inject_visible` now uses `write_two_phase_injection`: the labelled line, a 50ms settle gap (`INJECTION_SUBMIT_DELAY` const — deliberately not a config key, so no coupled binary/config rollout), then a lone `\r` as its own write. `last_local_input_at` is stamped after the final write so signal-less quiescence measures from the injection's end.

**#115 — dash-spawned workers never report back.** Two halves: `compose_worker_prompt` now logs a `report-back-omitted` decision event when mail is enabled but the requester id is unaddressable (previously silent); worker panes carry `report_to`/`report_reminder_sent`, and a new `report_back_reminder_sweep` injects a one-shot reminder naming the exact `zirv ctx send --to-session <id>` command into an idle worker pane that has produced output. The reminder is deliberately unconditional-once: no durable sender-side "already sent" signal exists (`mail-consumed` is logged by the recipient on read), so the wording makes a duplicate a harmless no-op.

v2.25.2. Vault updated (Ctx Supervisors, Known Issues, Decision Log, Work Journal, Active Work).

**Residual:** the live-codex behavioral proof (CR read as submit after the settle gap) needs the Docker AI-feature matrix run, which is operator-executed — unit tests pin the two-write shape and the once-only sweep semantics.


Glubiz:
## ✅ Ready to merge — CI green

All three checks pass on `0ac739b` ([run 32809798398](https://github.com/Glubiz/zirv-dynamic-cli/actions/runs/32809798398)):

| Check | Result |
| --- | --- |
| Test, Lint, and Format Check | pass (2m59s) |
| Serial Test Compatibility | pass (4m6s) |
| Version Bump Check | pass (8s) |

**Merge this PR before #117** — see the ordering note at the bottom.

---

### What this fixes

**#114 — pane injections were never submitted.** `zirv ctx nudge`, mail advisories and report-back injections wrote the visible line and its trailing `\r` in a single pty write. A codex TUI composer treats that burst as a paste and folds the `\r` into the text as a newline, so the message sat in the input box until a human pressed Enter. `Pane::inject_visible` now writes phase 1 only, stamps `injected_awaiting_turn`/`last_local_input_at` immediately, and schedules `pending_submit`; the dashboard tick drains it via `drain_pending_submits`/`Pane::submit_pending`, writing the lone `\r` as its own write. Operator keystrokes flush a pending submit first, and a failed CR write is retried rather than re-typing the line.

**#115 — dash-spawned workers never reported back.** `compose_worker_prompt` now logs a `report-back-omitted` decision event when mail is enabled but the requester id is unaddressable (previously silent), worker panes carry `report_to`/`report_reminder_sent`, and `report_back_reminder_sweep` injects one reminder naming the exact `zirv ctx send --to-session <id>` command into an idle worker pane that has produced output. Roster entries persist both fields (serde defaults keep old roster files parseable) so a restored pane keeps its report-back state; `Pane::handover` resets only the reminder flag, since a successor session should be able to be reminded once itself.

### Review history

Two review rounds ran against this branch. The first produced 7 findings, all fixed: the phase-2 failure double-type hazard and the UI-thread sleep burst (both resolved by the deferred-submit redesign rather than a patch), restored panes losing `report_to`, handover keeping a stale reminder flag, a duplicated addressability predicate, and a test helper that had drifted from the shipped path. A codex cross-review independently surfaced the same phase-2 hazard. A follow-up re-review of the fix commits returned no findings.

### One deliberate scope reduction

An earlier commit also applied the deferred submit to `wrap`'s `/compact` and mail-advisory injection. That broke two `cfg(unix)` real-pty tests, which timed out at CI's 180s hard limit rather than failing their own 15s budgets. Root cause: unlike a dash pane, the wrapped child only makes the advisory observable once its own line reader completes, which needs the CR **in-band** — and `read_until` checks its deadline only before each blocking `read()`, so a missing line hangs instead of failing fast.

Since #114/#115 are about dash panes and the wrap change came from a speculative review finding rather than an observed failure, `0ac739b` reverts the wrap half only. The dash-pane fix — the one the issues actually require — is untouched. Both wrap injection sites carry doc comments recording the residual, tracked in #118 along with the `read_until` diagnosability problem.

### Verification

- Local: build, `fmt --check`, `clippy --all-targets -D warnings` all clean; `commands::ctx::dash::` 317 passed / 0 failed; `commands::ctx::wrap::` failure-name set matches this machine's documented pre-existing Windows baseline exactly (the two T13 tests are no longer among them).
- CI: green, including the Linux PTY tests that cannot run on the dev machine.
- One CI run hit an unrelated flake in `commands::workflow::artifact::tests::a_server_command_that_exits_immediately_is_a_failure` (a stderr-capture race in a different module); re-running that job with no code change passed clean.

### Not verified here

The live behavioural proof — a real codex composer accepting the deferred `\r` as a submit — needs the Docker AI-feature command matrix, which is an operator-executed gate. Unit tests pin the two-write shape and the once-only sweep semantics; CI proves the PTY paths on Linux.

### Merge order

Merge **#116 first, then #117**. The version check compares against `main` at merge time: this PR lands 2.25.2 (main is 2.25.1) and #117 lands 2.26.0, which clears both. Merging #117 first would push `main` to 2.26.0 and make this PR fail its version check.

The source files auto-merge cleanly between the two branches. After this lands, #117 will need a small conflict resolution limited to `Cargo.toml`/`Cargo.lock` and five vault doc pages — no source conflicts.

PR #117, opened 2026-08-24
feat(safety): harden cross-harness permissions

Glubiz:
## Summary

- thread explicit `LaunchMode` through every real launch seam and preserve the fail-closed headless posture
- make interactive Claude low-noise: unknown/everyday work is silent, recoverable danger asks, and irreversible or credential-compromising effects deny
- apply one structural/semantic effect classifier to Bash, PowerShell, cmd.exe, executable paths, and Windows wrapper extensions
- attest each Claude launch with an immutable SHA-256 policy snapshot, keep the stricter launch/current verdict, and fail closed on tampering
- append privacy-preserving daily safety audit records containing command hashes rather than raw shell text
- retain Codex workspace containment with capability-probed interactive approvals/reviewer support
- bump Zirv to 2.28.0 and update the README plus architecture/security vault pages

## Cross-platform effect coverage

The semantic floor covers recursive generated cleanup, destructive Git effects, remote-versus-loopback HTTP mutations, credential-file access/upload, SQL writes, infrastructure/service destruction, and irreversible package/release operations. Ordinary reads, staging-only operations, downloads, loopback requests, dry runs, generated cleanup, and novel commands remain silent.

Native Windows Claude still lacks Claude's kernel sandbox. Landstrip, Arapuca, and Microsoft MXC were evaluated but not adopted as an invisible mandatory dependency because none currently proves a mature identical three-OS boundary. That residual is documented rather than presented as containment parity.

## Verification

- `cargo build` — passed
- `cargo fmt -- --check` — passed
- `cargo clippy --all-targets -- -D warnings` — passed
- `cargo test --bin zirv --no-run` — passed
- `cargo test --verbose -- --test-threads=1` — 2,409 passed / 5 failed; failures are exactly the established native-Windows `wrap` baseline:
  - `a_wrap_outside_any_session_is_not_gated`
  - `allow_nested_overrides_the_guard`
  - `the_allow_nested_environment_variable_overrides_the_guard_too`
  - `the_undetected_command_error_names_the_registry_rather_than_claude`
  - `win::a_supervised_wrap_binds_a_turn_signal_transport`
- all 88 `commands::ctx::safety` tests passed within the serial run
- `cargo nextest run --no-fail-fast` — unavailable on this host because `cargo-nextest` is not installed

## Review

Please use the orchestrator review path. Codex review is intentionally skipped per operator request.

## CI fix (2026-08-25)

- added the `interactive: false` field to the test-only `HandoverRequest` initializer in `dash/pane.rs` that the merge from `main` left behind (test builds failed to compile on both CI jobs)
- rewrote a `match` in `safety.rs` as `let end = end?;` to satisfy `clippy::question_mark` on current stable, which CI runs after tests


Glubiz:
## ✅ Ready to merge — CI green, conflicts resolved

All three checks pass on `feedcd4` ([run 32813657952](https://github.com/Glubiz/zirv-dynamic-cli/actions/runs/32813657952)):

| Check | Result |
| --- | --- |
| Test, Lint, and Format Check | pass (4m37s) |
| Serial Test Compatibility | pass (4m55s) |
| Version Bump Check | pass (6s) |

`main` (now 2.25.2 after #116 merged) has been merged into this branch, so `mergeable: MERGEABLE` — GitHub reports only the standard `BLOCKED` for a required review approval. Version lands at **2.26.1**.

---

### What this delivers

One zirv-owned permission policy, compiled into each wrapped harness's own native mechanism at launch. The design goal is the one that matters day to day: **only genuinely dangerous commands prompt.** An everyday dev command, a read-only command, or a command zirv has never seen before runs silently on an interactive launch — a finite allowlist under a prompting mode was explicitly rejected, because every novel command would fall off the end of it and prompt.

That requirement is pinned as executable tests, not prose:

- `the_product_requirement_no_everyday_or_novel_command_ever_prompts` — ~45 everyday and novel commands must classify Allow
- `the_product_requirement_only_genuinely_dangerous_commands_prompt` — the dangerous set must classify Ask
- `the_headless_posture_does_not_inherit_the_interactive_permissiveness` — headless stays fail-closed, because nobody is there to answer

Headless launches (`exec`, `run_loop`, scripted spawns) keep today's fail-closed posture throughout. Interactive Claude moves to `--permission-mode default` with the safety hook as the sole prompting gate; interactive Codex moves to `--ask-for-approval on-request` (the low-noise polarity — `untrusted` prompts for everything outside a narrow built-in set). Where Codex cannot match Claude's per-command granularity, `policy_support` reports `Degraded` honestly rather than claiming parity.

### Review history — three rounds, and the last one mattered

**Round 1** found 15 findings. All were verdict-direction bugs of the security-critical kind: commands that *should* prompt running with no prompt at all. Highlights: `curl url|sh` (no space) bypassing the pipe-to-shell deny; a crafted SQL comment (`SELECT '/*' ; DROP TABLE users ; SELECT '*/'`) mis-stripped into a single "read-only SELECT"; `is_local_url` matching any host starting `127.` so `127.evil.com` exfil was suppressed; `cargo +nightly publish` dodging the irreversible-publish deny; and non-tty `wrap`, hook-mode without `permission_mode`, and unwatched dash worker panes all receiving the *permissive interactive* posture instead of fail-closed headless.

**Round 2** was an adversarial re-review of those fixes — and it earned its keep. Seven fixes were confirmed closed, but four bypasses survived, including one shipped *inside* the fix that claimed to close it:

- **`awk` and `sed` were on the find-exec "proven-safe" allowlist.** Both are arbitrary-code-execution primitives (`find . -exec awk 'BEGIN{system("id")}' {} \;`, GNU `sed '1e id'`) — a complete escape from the ask-unless-proven-safe design.
- Pipe-to-shell missed wrapper programs: `curl x | env sh`, `| sudo sh`, `| xargs sh`.
- The SQL stripper was still erasable two other ways: MySQL backslash escapes, and PostgreSQL dollar-quoting (`psql -c "SELECT $$/* $$ ; DROP TABLE users; -- */"` classified Allow — a genuine unprompted DROP).
- `kubectl --context prod delete pod x` read `prod` as the verb and slipped past the destructive-orchestrator deny (only `-n`/`--namespace` were handled).

**Round 3** fixed all four: `awk`/`sed` removed and every remaining allowlist entry re-audited against a stated admission rule (a program qualifies only if it cannot execute another program or write outside its arguments); pipeline wrapper-unwrapping with a depth cap, handling `env`'s `VAR=value` prefixes and `timeout`'s duration positional; backslash-escape and dollar-quote modelling in the SQL scanner, with anything it cannot confidently parse falling back to Ask; and 21 realistic kubectl/helm global value-flags. Every fix carries the exact PoC as a corpus assertion, so each bypass is now a regression test.

Round 2 also specifically audited the two tests that an earlier round had "corrected in place" — the highest-risk move in the whole diff. Both genuinely encoded the bug as expected behaviour (`curl -G -d` expecting Allow; missing `permission_mode` expecting Allow) and were made **stricter**, not weakened. No assertion was deleted or relaxed anywhere in this PR.

### Known residuals (documented in code, not silently dropped)

1. **Symlink/junction in the generated-cleanup fast path** — `rm -rf node_modules` is auto-allowed by first path segment; a `node_modules` replaced with a symlink to a sensitive directory would be deleted through. Not fixed because the `evaluate` pipeline is deliberately pure with no filesystem access; a fix belongs caller-side.
2. **`claude.rs` `policy_support` cannot see whether `launch_settings_path` actually wrote** — if that best-effort settings file fails for one launch, that launch has no hook, while the report still says `Degraded`. Needs a signature change or an argv fallback.
3. **Codex capability probes are process-wide `OnceLock`s** — a codex-cli upgrade or downgrade mid-session keeps the stale first answer.

All three are defense-in-depth gaps rather than active bypasses, and all three carry doc comments at the relevant code.

### Not verified here

The operator gates from the spec remain outstanding: the Linux/Docker run beyond what CI covers, and the **Docker AI-feature command matrix** (`ctx chat/agent/dash/mail/hook`), which is harness-facing and operator-executed. Local Windows verification plus CI (both Linux jobs) is what backs the green above.

### Merge note

#116 is already merged (`main` at 2.25.2, tag `v2.25.2`), so the ordering constraint is satisfied and this can merge whenever you approve it.

PR #153, opened 2026-08-26
fix(ctx): codex approval posture — config-override pinning, writable roots, network knob

Glubiz:
Companion to #147's approval-hell investigation, codex side. Version 2.30.0 → 2.32.0.

Implements and gates config-override pinning, writable roots, and network knob for codex approval posture:
- `flags_pin_policy` now recognizes operator `-c approval_policy=…`/`-c sandbox_mode=…` (both split and joined forms) so zirv never appends its own posture args after an operator override.
- `extra_writable_root_args`: dash-spawned codex panes get one merged `-c sandbox_workspace_write.writable_roots=[…]` naming the linked-worktree git common dir and the zirv state mail dir (report-back delivery).
- New operator-owned `[policy] network` knob → `-c sandbox_workspace_write.network_access=true` for codex when allowed.
- `permissions audit/compile --agent codex` prints a caveat that compiled `[safety] allow` entries do not affect codex launches.

## Status
All review findings fixed with named regression tests (d00f663); CI green; ready for approval. The PR is undrafted and mergeable.

PR #154, opened 2026-08-26
fix(ctx): sandbox-escape allowlist ends Claude approval hell

### LINKED ISSUE #147: fix(ctx): Claude sandbox retries ignore standing allows and repeatedly reprompt
## Summary

Zirv-managed interactive Claude sessions repeatedly request permission for
commands already allowed by operator policy. PR #142 supports Claude audit
compilation, but ordinary `[safety] allow` rules cannot satisfy Claude's
separate sandbox-escape/native permission layer.

## Production evidence — 2026-08-25/26

- 152 Claude `Ask` decisions across the two latest state-log buckets.
- 123/152 were `<sandbox: unsandboxed retry>`.
- Transcript-hash correlation found approximately 50 cargo verification asks,
  25 shell/file-probe asks, 16 `gh issue`/`gh pr` asks, and 12 Zirv CLI asks.
- Those families already have `cargo *`, `gh *`, `zirv *`, and common
  read-only shell allowances.
- `permissions audit --agent claude --sessions 50` reported 15 denials:
  `find` 7, `cd` 5, `grep` 2, `ls` 1.
- A correlated `grep`/absolute-path `find` invocation was allowed by the Zirv
  hook and then denied by Claude's native permission rules.
- The operator's Claude settings contain `Bash(*find /*)`, which overmatches
  safe `/Users/...` paths as well as root searches.
- Some audited compound commands contain `rm -rf`, root-wide `find`, or `.env`
  access; compiling only their first-token family would be unsafe.

## Root cause

The safety hook deliberately converts every nontrivial sandbox escape to Ask,
regardless of an existing semantic Allow. The launch settings also inject a
native `Bash(dangerouslyDisableSandbox:true)` ask, and Claude applies the most
restrictive merged verdict. Main's audit records only errored tool results and
mislabels all of them as `dontAsk`, while accepted interactive prompts are
absent.

PR #142 is not Codex-only, and its output is reusable family wildcards rather
than full argv. However, it writes ordinary home-level safety allows, rejects
the single-token families above, and does not compile or project a Claude
sandbox-escape approval. Not writing allow rules into repo settings is correct:
repo-owned configuration may only narrow policy.

## Acceptance criteria

- Repeated eligible Claude sandbox asks can be explicitly compiled into a
  separate operator-owned escape allowance; repo configuration cannot set it.
- Equivalent `cargo test`/`cargo nextest`, read-only `gh issue`/`gh pr`, and
  `zirv ctx` commands stop prompting after operator approval.
- Root-wide searches, `.env` access, `rm -rf`, and destructive VCS operations
  continue to ask or deny.
- Compound commands are classified across every shell segment; `cd && rm -rf`
  can never produce a reusable `cd` allowance.
- Audit reports actual mode, sidechain status, native denial versus sandbox ask,
  and includes accepted prompts through decision-log correlation.
- Stronger native Claude deny rules are surfaced as conflicts instead of
  producing ineffective allow recommendations.
- Interactive and headless regression tests cover hook failure, attestation,
  native precedence, sandbox escape allow, ask, and deny behavior.

## Fix plan (proposed)

1. Extend PR #142’s compiler to correlate Claude transcripts with safety-decision hashes and classify actual asks, denials, permission mode, and sidechains.
2. Add a separate operator-owned, repo-forbidden sandbox-escape allowlist; compile only reusable multi-token families after validating every compound-command segment.
3. Let an unsandboxed retry proceed only when semantic policy returns `Allow` and the escape allowlist also matches; retain current interactive Ask/headless Deny otherwise.
4. Replace the unconditional native sandbox ask with the attested hook decision, failing closed when the hook or policy attestation is unavailable.
5. Detect stronger native deny conflicts such as `Bash(*find /*)` and offer a backed-up migration only for known Zirv-managed legacy entries.
6. **Base verdict: build on PR #142**, because its operator-owned compiler is the correct foundation; main alone would duplicate that open work.

*(Investigation by codex worker, dash pane 1462a245.)*

Glubiz:
Fixes #147. Version 2.30.0 → 2.31.0.

Implemented (all five gates green at the last full run, zero new test failures by name-diff vs main):
- Operator-owned, repo-forbidden `[safety] escape_allow`: an unsandboxed retry keeps Allow only when attestation is OK, semantic policy allows, and every shell segment matches the escape list.
- Built-in escape seed for read-only utilities (grep/cat/ls/find/…) with a credential-path/root-find/.env screen — covers the observed read prompts (71% of interactive asks were unsandboxed retries).
- Removed the injected native `Bash(dangerouslyDisableSandbox:true)` ask rule: per documented Claude Code precedence, hook allows never bypass settings ask rules, so it silently defeated every hook-side allow. The attested fail-closed hook is now the sole zirv-side escape gate.
- Claude sandbox now write-allows the state mail dir so `zirv ctx send` report-back works in-sandbox (pairs with #151's EPERM fix).
- `permissions audit` correlates the safety decision log (true cause/mode/sidechain, accepted prompts); `compile --escape` writes eligible multi-token families to escape_allow (home layer only); native deny/ask conflicts surfaced instead of ineffective recommendations.
- Codex compile caveat and `-c` pinning live separately in #153.

## Handover — draft until final polish
- HEAD (b6b7373) is an orchestrator checkpoint snapshot: it captured a vault-keeper subagent mid-run. Command Safety / Ctx Adapters / Ctx Subsystem / Untrusted Configuration / Decision Log vault pages are updated; **Work Journal + Active Work are not yet**.
- A doc-comment-only fix after the last full gate run was verified with build/fmt/clippy but not re-run through nextest (no logic touched).
- Before merge: check for trailing vault-keeper edits in the worktree, update the two missing vault pages, run all five gates once end-to-end, squash the wip commits into a clean history.


Glubiz:
**Handover status: NOT READY — security fix round in progress.** The rebased, squashed head `b15e8f8` passed all five local gates and delivers the escape allowlist, the interactive-mode pin (kills the defaultMode:"auto" ask-everything flood), the CI artifact.rs race fix, and vault docs. The independent review gate then found one Critical: `escape_denied_by_screen` never checks shell output redirection, so a seeded "read-only" family (`echo`, `cat`, …) retried unsandboxed can silently write to arbitrary paths; compounding, `SpawnRequest.interactive` is trusted verbatim from the drop directory, letting a forged request self-escalate a new pane's mode pin. A fix round is running now: route the screen through the existing `contains_unquoted_redirection` primitive (+ the permissions `--escape` compile path), bind the interactive pin in-process so file-dropped spawn requests can never carry it, with security tests for both. Stays draft until the fix passes a scoped re-review.

Glubiz:
**Shutdown handover (session end).** The security fix round is DONE locally: commit `b46915d` (screens unquoted redirection in `escape_allow` + the permissions `--escape` compile path, binds the interactive spawn-pin in-process so file-dropped requests can never carry it, with security tests). It could NOT be pushed onto this PR branch: the history was amended and the operator's own deny rule blocks `git push --force*` while unattended — by design. The identical head is pushed as branch **`fix/147-sandbox-escape-allowlist-fixround1`**.

Morning steps: (1) `git push --force-with-lease origin b46915d:fix/147-sandbox-escape-allowlist` (or reset the PR branch to the side branch); (2) scoped re-review of the fix vs the Critical/Important findings above; (3) full nextest name-diff; (4) undraft. Full details in `D:\GitHub\zirv-pr154\PR154-REPORT.md` (fix-round section appended).

Glubiz:
## Status: READY TO MERGE — position 1 of 6, head of the train

**Blocking on:** one approving review (the "Protected Main" ruleset requires 1). Nothing else.

### State at handover (head `a6a0bb4`)
- CI: all three checks green.
- Version: 2.31.0.
- Security review loop closed after three fix rounds; no blocking findings remain.

### Work completed this session

**1. CI was red on a broken test — fixed, and it was not a regression.**
`a_forged_interactive_spawn_request_still_spawns_normally` does a real pty spawn but never overrode `cfg.agent_bin`, unlike every other real-pty-spawn test in `dash/mod.rs` (which pin it to `sleep 3` / `ping -n 3 127.0.0.1` per the convention documented right beside them). It resolved to a bare `claude` on PATH, so it passed by accident on machines with a real `claude` installed and failed on a clean runner. Fixed by pinning `agent_bin` in the test. No production code touched; the security property was never affected.

**2. A CRITICAL bypass found by review — closed over two rounds.**
`is_root_wide_find_scan` (`src/commands/ctx/safety.rs`) only inspected `tokens.get(1)` for `/`, `~`, `~/`. Any leading option shifted the real starting point past that fixed index, so `find -H / -iname id_rsa` rode the seeded `find *` escape_allow family to a **silent unsandboxed Allow with no prompt** — an unauthenticated root-wide scan for credential-named files. It was reachable at hook time and, through the shared `escape_denied_by_screen` path, at `permissions compile --escape` time too. This directly violated issue #147's own criterion that root-wide searches continue to ask or deny.

- *Round 1* rewrote the check to scan every token (mirroring `is_risky_find_exec`'s "trust no fixed position" stance) and added a `$`/backtick disqualification so `find $(echo /)` can't smuggle a root through command substitution.
- *Round 2* replaced exact string matching with **lexical normalization**, because round 1 still let through `//`, `/.`, `/./`, `/..`, `/../`, and `~root` — all verified as live silent allows. Exact-match against a set of spellings could never converge; `resolve_lexical_path_components` now collapses slashes and resolves `.`/`..` textually (never walking above its start, no filesystem access, so the classifier stays pure), and `is_root_wide_or_whole_home_path` treats any bare `~`/`~user` as unbounded.

Eight regression tests, each confirmed failing before the fix, plus a positive case proving bounded paths (`find /home/jonathan/project -name x`) still escape — no over-denial. Independent review verified all ten PoCs now ask, tried `/foo/../..`, `~/..`, `~/../..`, `~root/..` without finding a hole, and confirmed no bounded path regressed.

**3. Both earlier-round security fixes were re-reviewed and are SOUND** — the `escape_denied_by_screen` unquoted-redirection screen, and the `trusted_interactive` / `FILE_DROP_TRUSTED_INTERACTIVE` binding (`req.interactive` from file-drop JSON is provably never read by `spawn_launch_mode_pin`).

### Issue #147 acceptance criteria
All met. Native-deny precedence is sound by construction (Claude ANDs hook and native settings independently, so a zirv allow cannot beat an operator's native deny). Audit correlation joins on session + command_sha256 within a 300s window and falls back to generic wording rather than fabricating a match.

### Residual findings — tracked in #160, none blocking
1. Restored dashboard panes never receive the interactive launch-mode pin (fails closed, so safe direction; pre-existing).
2. `is_root_wide_find_scan` still ignores relative paths (`find ..`) — a documented non-goal of this text-only classifier, never in scope for these rounds.
3. zirv's own safety layer denied a `git commit` heredoc because the *message text* contained `~/` and `/../`.

### Merge order
`**#154 (2.31.0)** → #153 (2.32.0) → #156 (2.33.0) → #157 (2.34.0) → #158 (2.35.0) → #159 (2.36.0)`

This goes first: it shares six files with #153 and touches `dash/mod.rs` (+185) and `permissions.rs` (+1234) heavily, so #153 rebases onto it rather than the reverse.

### Next session picks up
Approve and merge, then rebase #153. No code work outstanding.

Glubiz:
## Do not merge separately — folding into a single combined release

Per operator decision, all six open PRs are being combined into one integration branch (`release/2.31.0`) so the project ships **one release tag** instead of six. CD fires on every merge to `main`, so only the integration PR may merge there.

This PR has been set back to draft purely to prevent an accidental separate merge. **Its work is not cancelled** — every reviewed commit is carried onto the integration branch as-is, and the review history and handover comment above remain the record for this slice of the work.

The per-PR version bumps (2.31.0–2.36.0) are superseded by a single `2.31.0` on the integration branch.

This PR will be closed once the integration PR merges. If the integration turns out not to be viable, it can be marked ready again and merged on the original ladder.

PR #156, opened 2026-08-26
perf(ctx): loud truncation, trimmed common.md, one memory layer at the cacheable tail (#155 phase 1)

Glubiz:
Phase 1 of #155's token-cost-reduction plan (quick wins). Three levers, each measured and test-pinned:

1. **Loud truncation** — a canonical `.zirv/context/` layer that gets cut now writes a `context-truncated` decision-log entry (file, delivered/raw/lost bytes, the operator config key) plus an unconditional stderr note at compose time, gated by an explicit `log_truncation` parameter so read-only `context status` compiles stay silent. Before this, 821 bytes of this repo's own instructions were being cut mid-word, silently, every session.
2. **`common.md` fits its budget** — trimmed 4917 → 3898 bytes (every rule preserved in meaning; rule-by-rule reviewed), with repository-content pin tests so the file can never silently outgrow the 4096-byte cap again.
3. **One memory layer, at the cacheable tail (v7 → v8)** — the composed prompt previously carried TWO memory layers, and the git-diff-driven retrieval layer sat before the canonical context layer, re-pricing ~8KB as uncached input on essentially every recompose during active editing. Now a single deduped memory layer is injected after the canonical context layer, so the cache-stable prefix ends only where genuinely per-session content begins.

Version 2.33.0 (2.31.0/2.32.0 are reserved by PRs #154/#153 — merge those first).

## Verification

All five gates green on the PR head: `cargo build`; `cargo nextest run --no-fail-fast` (2601/2602 — the single failure is this dev machine's known `wrap::win::turn_signal` baseline name); `cargo test --verbose -- --test-threads=1` (same single baseline name); `cargo fmt -- --check`; `cargo clippy --all-targets -- -D warnings`. Every task passed an independent review gate, and a whole-branch final review returned MERGE-READY with no findings:

> Whole-branch review found full cross-task coherence, no spec gaps, and no new config keys or behavior changes for non-zirv-launched sessions. Two pre-ledgered documentation minors are correctly left deferred.

Part of the #155 train: telemetry (P2), context dedupe (P3), review convergence (P4), work groups (P5), model-aware rotation (P6) follow in separate PRs.

Issue #155, opened 2026-08-26
perf: make zirv sessions dramatically cheaper to run without losing response quality


Glubiz:
Production evidence, 2026-08-26 (this session): one orchestrated fix round (3 branches, ~10 workers) drove the five-hour usage window from 10% to 95%. Observed worker costs: implementers 270k-692k output tokens each (408 tool uses on the largest), investigators/reviewers substantial on top. zirv itself contributes per-session cost: the v7 prompt injection (default+adapter+harness+roster+memory+user+canonical context) is re-composed per session, and mail/nudge/report-back injections add mid-session payloads.

Goal: materially reduce token cost per unit of completed work while holding review-verified quality (finding rates, gate pass rates) constant.

Candidate work items (each measurable on the decision/usage logs zirv already keeps):
1. **Cache-stable injection**: keep every injected layer byte-stable across turns and sessions (stable ordering, no timestamps/session ids inside the cached prefix) to maximize provider prompt-cache hits; measure cache-hit deltas.
2. **Injection diet**: per-verb minimal prompts — headless workers don't need the full dashboard/mail/roster layers; canonical context and memory entries capped and deduped (hash-skip unchanged layers on restart/handoff).
3. **Worker tiering enforcement**: default worker model tier per task class (mechanical/ordinary/hard) enforced by `zirv agent`, with the orchestrator seat reserved for coordination; warn when a worker inherits an expensive seat model.
4. **Per-worker token budgets**: extend the heavy-worker slot budget (#142) with a token budget per delegation — refuse or downshift model when exceeded; surface spend per worker in `zirv ctx usage`.
5. **Review-train cost control**: single-reviewer default, effort caps, and re-review scoped to fix diffs only (a single high-effort fan-out previously consumed ~48 points of a five-hour window).
6. **Compaction/handoff earlier on cost signals**: let the rot engine treat token spend rate as an input so long sessions hand off before the window exhausts (usage/pace data already exists).
7. **Telemetry to prove it**: per-verb and per-worker token accounting in the decision log so each of the above lands with a before/after number.

Acceptance: a documented baseline (tokens per completed task from the logs), each shipped item showing its measured reduction, and no regression in review-confirmed defect rates on the same workload.

Glubiz:
**Epic status (2026-08-27, autonomous overnight run).**
[...]

PR #157, opened 2026-08-27
feat(ctx): raw four-category token telemetry with delegation lineage (#155 phase 2)

Glubiz:
**Status: DRAFT — 4 of 6 phase tasks complete, review-gated; 2.4 in flight.** Phase 2 of #155 (P0 measurement): telemetry stops collapsing Anthropic's four raw token categories and gains delegation lineage, so every later optimization lands with a before/after number.

Landed, each through an independent review gate + fix rounds:
- **2.1** (`a6e1319`) — `TranscriptUsage` carries {input, cache_creation, cache_read, output} raw; adapters stop pre-summing; `context_total()` preserves every legacy combined consumer byte-for-byte.
- **2.2** (`129133c`, `41fbdf4`) — `TelemetryEvent` v2: session lineage + a sidechain bucket; native child-agent spend is now counted (previously invisible to workflow accounting while still hitting the 5h quota).
- **2.3** (`6d4813c`, `a0b5105`) — every completed `zirv agent` delegation writes a `Delegation` record (model, four raw categories, wall time, outcome) to `delegations.jsonl` + a `delegation-complete` decision.

In flight: **2.4** `zirv ctx usage --sessions` (per-session raw categories + cache-hit ratio, last 24h), then 2.5 version bump (renumbered to 2.34.0 at stack time) and 2.6 phase gates.

Semantics ruling baked in: cache-hit ratios use RAW fields only; `TelemetryEvent.input_tokens` keeps its legacy combined meaning and never enters a ratio.

Stacked on #156 (base = `feat/token-cost-p1-quick-wins`). Merge order: #154 → #153 → #156 → this.

PR #158, opened 2026-08-27
feat(context): canonical-content hash stamps toward native dedupe (#155 phase 3)

Glubiz:
**Status: DRAFT — 1 of 4 phase tasks complete (under review); the dedupe skip itself is next.** Phase 3 of #155: stop injecting ~8KB of canonical `.zirv/context/` into every composed prompt when the harness already reads the same content natively from the generated `CLAUDE.md`/`AGENTS.md`.

Landed:
- **3.1** (`d825b45`) — `render_generated` stamps `<!-- zirv:canonical-sha256:… -->` (length-prefixed domain separation) into the managed prefix of generated harness files; reader helper ready for the skip check; this repo's own CLAUDE.md/AGENTS.md regenerated with stamps. Review in progress.

Next: **3.2** — compose-time skip: when the harness-native target exists and its stamp matches the current canonical sources, omit the canonical injection (`context-dedup-skip` decision; `context.dedupe_native` config, repo layer may only narrow to `false`); **3.3** version bump (renumbered to 2.35.0 at stack time); **3.4** phase gates.

Stacked on #156 (base = `feat/token-cost-p1-quick-wins`; will restack onto Phase 2 when it lands per the plan's sequencing). Merge order: #154 → #153 → #156 → P2 → this.


Glubiz:
**Shutdown handover (session end).** Task 3.2 (the compose-time dedupe skip) was mid-implementation at the stop order and did not reach a green checkpoint — nothing was committed beyond 3.1. Its uncommitted WIP (compile.rs, config.rs, context_cli.rs; ~15KB) is snapshotted at `zirv-155-p3/.superpowers/sdd/2026-08-26-token-cost-reduction/task-3.2-wip-snapshot.diff` on this machine. Resume by re-dispatching Task 3.2 with its brief + the snapshot as reference (or from scratch — the brief is self-contained). Then 3.3 (bump to 2.35.0) and 3.4 (gates).

Glubiz:
## Status: COMPLETE — all 4 phase tasks done, reviewed clean. Position 5 of 6.

**Blocking on:** #156 merging first (this is stacked on it), then undraft + one approving review.

### State at handover (head `6c234d7`)
- All five gates green: build, `cargo nextest run --no-fail-fast`, `cargo test --verbose -- --test-threads=1`, `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings`. Two unrelated `exec.rs` flakes appeared once each under host contention and passed on isolated and full re-runs — this machine's known pattern, not regressions.
- Version: 2.35.0.

### Task 3.2 — the compose-time dedupe skip (the phase's core deliverable)

The previous session's WIP for this was lost with a deleted worktree, so it was reimplemented from the epic spec rather than recovered.

`compile.rs::with_canonical_context_layer` now consults `native_file_already_carries_canonical(adapter_name, repo, cfg)`, which proves the adapter's own `CLAUDE.md`/`AGENTS.md` already holds the exact canonical bytes. When it does — and `cfg.context.dedupe_native` (new, default `true`) is set — every candidate is still read and still reported in `ContextProvenance` (`delivered_bytes: 0, truncated: false`, so `zirv context status` still sees it), but nothing is appended and `PromptSource::Context` is omitted. One `context-dedup-skip` decision is logged under the same gate truncation logging uses.

**A bug in the plan document's own pseudocode was caught during implementation:** it suggested reusing `narrow_pace_bool` for the new config key, which has the wrong polarity — a repo-absent value would have silently forced `false` everywhere and disabled the feature by default. A correctly inverted `narrow_dedupe_bool` was written instead. Review verified all four operator/repo polarity combinations: a repo layer can only narrow toward `false` (inject *more*) and can never re-enable a skip the operator disabled. `context.dedupe_native` is deliberately **not** `REPO_FORBIDDEN` for exactly that reason, so the README trust-table bijection test correctly does not require a row for it.

### Review round: 1 critical + 1 major, both fixed and re-verified

**CRITICAL — the proof trusted a self-declared receipt.** The original check extracted the embedded hash from the native file's header comment and compared it against a fresh hash of the *current `.zirv/context/` sources* — it never hashed the native file's own **body**. So generating `CLAUDE.md` correctly and then hand-editing only the body, leaving the header intact, still satisfied the check: dedupe fired and the session received **none** of the real canonical text. The reviewer confirmed this with a tamper probe. This was precisely the failure the module's own doc comment says must not happen.

Fixed: the header hash is now a cheap pre-filter only. The proof is byte-exact `native_text == context_cli::render_generated(common, harness)` against the file's actual current bytes, with no normalization — CRLF and trailing-whitespace differences count as real differences and fall back to injecting. A header/pre-filter mismatch can only ever cause a *non*-skip, never a false skip.

**MAJOR — deduped surfaces were indistinguishable from mysteriously-empty ones.** `zirv context status` only flagged `[TRUNCATED]`, so a skipped layer printed "500B raw, 0B delivered" with nothing explaining why. Added a `[DEDUPED]` marker, checked *after* `[TRUNCATED]` so a genuine zero-budget truncation is never misreported.

Four new tests (three confirmed failing pre-fix): tampered body behind an intact header, shortened/lengthened body, CRLF-only difference, and the status-marker case. Independent re-review re-ran the original tamper probe against the fix and confirmed injection now fires with the real text reaching `composed.text`, verified the happy path still dedupes for the right reason rather than vacuously, and found no new defects.

### Not measured
The `zirv context sync --generate` / `zirv ctx usage --sessions` cache-hit measurement from the plan doc was not run — it's a manual measure-and-record step, not a gate. Epic #155's acceptance asks for a before/after number per shipped item, so this is worth capturing before the epic closes.

### Merge order
`#154 → #153 → #156 → #157 → **#158** → #159`

Stacked on `feat/token-cost-p1-quick-wins` (#156). When that merges, retarget this to `main` and re-check CI.

### Next session picks up
After #156 and #157 land: retarget, undraft, approve, merge. Optionally record the cache-hit measurement. No code work outstanding.

Glubiz:
## Do not merge separately — folding into a single combined release

Per operator decision, all six open PRs are being combined into one integration branch (`release/2.31.0`) so the project ships **one release tag** instead of six. CD fires on every merge to `main`, so only the integration PR may merge there.

This PR has been set back to draft purely to prevent an accidental separate merge. **Its work is not cancelled** — every reviewed commit is carried onto the integration branch as-is, and the review history and handover comment above remain the record for this slice of the work.

The per-PR version bumps (2.31.0–2.36.0) are superseded by a single `2.31.0` on the integration branch.

This PR will be closed once the integration PR merges. If the integration turns out not to be viable, it can be marked ready again and merged on the original ladder.

Glubiz:
Superseded by #161, which merged as v2.31.0 and carries every commit from this branch. Closing as shipped.

PR #163, opened 2026-08-27
feat(ctx): work groups and `zirv ctx group create|status|close` (#155 task 5.2)

Glubiz:
Task 5.2 of epic #155, phase 5. Adds the persisted unit a delegated batch is accounted against.

`zirv ctx group create <scope> [--child-limit N] [--token-budget N] [--deadline-secs N] [--completion-contract "..."] [--parent-session ID]` mints a `wg-<uuid>`, persists it, and prints the id. `zirv ctx group status [id]` shows one group or all of them; `zirv ctx group close <id>` stamps `closed_at` and is idempotent, with the first close time always winning.

State lives at `<state>/groups/<id>.json`, one file per group, written through the same `create_private_dir_all` + `write_private` 0600 atomic-write convention `sessions.rs` uses. `token_budget`, `deadline_secs` and `closed_at` carry `#[serde(default)]` so older and newer files still parse, and `list()` skips an unparsable file rather than failing the whole listing or deleting it.

**Deviation from the plan, deliberate.** The plan says to add `groups()` to whatever `StateDir::ensure()` creates, "matching `sessions()`". On current main `sessions()` is *not* created by `ensure()` — only `handoffs()`, `sockets()` and `logs()` are — and `sessions.rs` creates its directory lazily at write time instead. Following the plan literally would have made groups inconsistent with the exact sibling it pointed at, so this follows the real precedent: lazy creation on write.

Nothing else consumes groups yet. `work_group_id` remains `None` at every call site until Task 5.3 wires spawn lineage.

Gates: build, `cargo nextest run --no-fail-fast` (2744/2744), `cargo test -- --test-threads=1` (2744/2744), `fmt --check`, `clippy -D warnings` — all clean.

Version 2.33.0 is provisional; the remaining phase-5/6 branches will be collapsed to a single version at release integration, as v2.31.0 was.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2693,7 +2693,7 @@ checksum = "b97154e67e32c85465826e8bcc1c59429aaaf107c1e4a9e53c8d8ccd5eff88d0"
 
 [[package]]
 name = "zirv"
-version = "2.31.0"
+version = "2.33.0"
 dependencies = [
  "clap",
  "console",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zirv"
-version = "2.31.0"
+version = "2.33.0"
 edition = "2024"
 authors = ["Jonathan Solskov <josj@zirv.io>"]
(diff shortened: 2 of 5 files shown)


Glubiz:

PR #164, opened 2026-08-27
docs: token-cost benchmarks document and measurement protocol (#155 task 6.4)

Glubiz:
Task 6.4 of epic #155. Docs-only, so it cannot conflict with the code branches in flight.

Issue #155 closes on measurement: "a documented baseline (tokens per completed task from the logs), each shipped item showing its measured reduction, and no regression in review-confirmed defect rates." This adds `docs/benchmarks/token-cost.md` defining how that is measured and recording what is actually known today.

**Every phase row currently reads "not yet measured."** That is the honest state, not an oversight. Verified on this machine:

- `delegations.jsonl` does not exist anywhere, despite the write path being wired in `agent.rs`. The sessions visible in `zirv ctx usage --sessions` were spawned as native harness Agent/Task-tool sessions rather than through the `zirv ctx agent` CLI, so that accounting path has never fired.
- `zirv workflow stats` reports 0 events and no `workflow-telemetry/` directory exists — `zirv workflow` has never been run here, so review-count-per-change and workflow-id task grouping have no data at all.

Each row therefore carries a specific verified reason and the exact command that will fill it once data exists. No number in this document was estimated, extrapolated, or invented; fabricated figures here would corrupt the epic's own acceptance evidence.

The one live source, `zirv ctx usage --sessions`, is captured as a clearly-labeled appendix snapshot and deliberately kept out of the results table: it is cross-repo, 24-hour aggregated, uncontrolled, and contaminated by the concurrent agents of the very session that captured it.

**One trap the document records.** Two token-record shapes share field names with different meanings: `TranscriptUsage` (and `delegations.jsonl`) has `input_tokens` as raw uncached with the cache fields additive, while `TelemetryEvent` has `input_tokens` as the already-combined total with the cache fields as subsets. Each gets its own cache-hit formula. Using the wrong one silently produces a wrong ratio — the same defect class caught in review on #157.

The protocol section pins model ids, effort, review policy, concurrency and repo state, per the architect's requirement on #155 that a model or settings change can never be reported as an efficiency win, and records median and p95 per completed task alongside the five-hour delta, cache-hit ratio, tool calls, wall time and completion rate.

Version 2.34.0 is provisional and will be collapsed at release integration.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2693,7 +2693,7 @@ checksum = "b97154e67e32c85465826e8bcc1c59429aaaf107c1e4a9e53c8d8ccd5eff88d0"
 
 [[package]]
 name = "zirv"
-version = "2.31.0"
+version = "2.34.0"
 dependencies = [
  "clap",
  "console",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zirv"
-version = "2.31.0"
+version = "2.34.0"
 edition = "2024"
 authors = ["Jonathan Solskov <josj@zirv.io>"]
 description = "A CLI tool for executing developer-defined YAML scripts with support for interactive commands and OS-specific execution."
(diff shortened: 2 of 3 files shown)


Glubiz:

PR #165, opened 2026-08-27
feat(prompt): make an active workflow review gate the single review source of truth (#155 task 4.1)

Glubiz:
Task 4.1 of epic #155 — the last missing piece of phase 4. Phase 4's engine half (delta re-review, stop-on-no-new-findings) shipped in v2.31.0; this is the prompt half that was deliberately deferred as stack-time work.

**The problem.** Three sources independently demanded a review round and none knew about the others: the meta-harness prompt layer, the claude adapter's orchestrator layer — which explicitly stacked itself *on top* of the first — and the workflow engine's risk-based reviewer count. A Medium-risk change was therefore reviewed three times over the same full diff.

**The fix.** Where a `zirv workflow` review gate is active, it is now named as the single source of truth in both prompt layers.

- `HARNESS_PROMPT` (`prompt.rs`): version token `(v9)` → `(v10)`; the final review bullet now opens by deferring to an active workflow gate and says not to run an additional native or cross-harness round on top of it. Everything else in that bullet is preserved verbatim — capacity-limited harnesses, triage, re-review only what the fixes touched, stop-on-no-new-findings, and the two-fix-round hard stop.
- `ORCHESTRATOR_PROMPT` (`adapters/claude.rs`): the sentence that caused the stacking is replaced by one deferring to the workflow gate, with the native review not running at all in that case.

The separate cost control immediately preceding it — route to the review model named in the harness roster, never the seat's own model, never a high-or-above fan-out — is untouched.

**Two assertions outside the plan's own test list needed updating**, both found by grepping rather than by the plan:

- Two existing tests asserted `HARNESS_PROMPT.starts_with("zirv meta-harness (v9)")`.
- `ORCHESTRATOR_PROMPT.len() < 3_000` began failing at 3,013 bytes with the first draft sentence. Rather than loosen a prompt-size budget — in an epic whose entire purpose is cutting token cost — the new sentence was trimmed to reuse `HARNESS_PROMPT`'s own "`zirv workflow review run` is the round" phrasing, landing back under the cap.

Gates: build, `cargo nextest run --no-fail-fast` (2742/2742), `cargo test -- --test-threads=1` (2742 passed, `test result:` line confirmed present so it was not a silent crash), `fmt --check`, `clippy -D warnings` — all clean.

Version 2.32.0 is provisional; the phase-5/6 branches will be collapsed to a single version at release integration, as v2.31.0 was.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2693,7 +2693,7 @@ checksum = "b97154e67e32c85465826e8bcc1c59429aaaf107c1e4a9e53c8d8ccd5eff88d0"
 
 [[package]]
 name = "zirv"
-version = "2.31.0"
+version = "2.32.0"
 dependencies = [
  "clap",
  "console",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zirv"
-version = "2.31.0"
+version = "2.32.0"
 edition = "2024"
 authors = ["Jonathan Solskov <josj@zirv.io>"]
 description = "A CLI tool for executing developer-defined YAML scripts with support for interactive commands and OS-specific execution."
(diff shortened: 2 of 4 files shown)


Glubiz:

Issue #170, opened 2026-08-27
Sub-orchestrator model: one layer below the meta orchestrator, each owning a scoped area bound to its own work group


Glubiz:
Operator-defined target model for sub-orchestrators and work groups, so the vocabulary shipped by epic #155 phase 5 grows into an actual delegation architecture rather than staying plumbing.

## The intended model

- **Exactly one layer of sub-orchestrators below the meta orchestrator.** The meta orchestrator (the expensive seat) decomposes the overall job; sub-orchestrators never spawn further sub-orchestrators or coordinators — only workers. Depth is fixed: meta → sub-orchestrator → worker, never deeper.
- **Each sub-orchestrator owns one bigger task with a coherent scope** — frontend, backend, a single service, or an equivalent natural boundary. The scope is the unit of delegation: the meta seat hands a sub-orchestrator a whole area, not a micro-task, and the sub-orchestrator splits it into worker-sized pieces at the cheapest model tier that fits.
- **One work group per sub-orchestrator scope.** The group (from `zirv ctx group create <scope>`) is the accounting and lifecycle unit: child limit, token budget, deadline, completion contract, and the per-child spend tree in `zirv ctx status` all hang off it.

## What already exists (shipped on the phase-5/6 release branch, 2026-08-27)

- `PromptRole::SubOrchestrator` with its own trimmed prompt layer: delegation vocabulary, no `HARNESS_PROMPT`, no harness roster, and an explicit prohibition on spawning another coordinator (task 5.1).
- Work groups persisted under `<state>/groups/`, with `zirv ctx group create|status|close` (task 5.2).
- Spawn-request lineage (`role`/`parent_session`/`work_group_id`) and a depth cap that already refuses SubOrchestrator→SubOrchestrator and anything below Worker (tasks 5.3/5.4). The single-layer rule is therefore already *enforced at the gate* — what's missing is the rest of the model around it.

## What's missing

1. **Role persistence** — the known gap from #169: nothing persists a pane/session's role, so a legitimately-spawned sub-orchestrator's own delegations currently read as Worker and are refused (fails closed). The single-layer model cannot function until `Pane`/`sessions::Record` carries the role it was spawned with, server-side, forgery-proof.
[...]

Issue #169, opened 2026-08-27
Follow-ups from 5.3/5.4: persist pane roles for sub-orchestrator chaining; accumulate worker budgets across restarts


Glubiz:
Two deliberate scope limits from Task 5.3/5.4 (branch `feat/token-cost-p5-lineage-budgets`), flagged by the implementer rather than silently dropped. Neither blocks the phase-5/6 release; both need follow-up work.

## 1. Sub-orchestrator chaining fails closed because no role is persisted

The depth cap derives a spawn request's parent role via `parent_role_for(req, panes)`: it checks only whether `req.parent_session` matches a pane in the dashboard's own in-process list — trusted server-side state the request cannot write to, following the `trusted_interactive` discipline. A match reads as Worker; no match reads as Orchestrator.

This is the correct forgery-proof shape, but it has a known over-restriction: a *legitimately spawned* SubOrchestrator pane's own further delegation also reads as Worker (and is refused) rather than as SubOrchestrator, because **nowhere in the codebase is a role persisted** — `PaneSpec::role` is read and discarded by `Pane::spawn`, and `sessions::Record` has no role field. The plan's "look the role up in the registry" instruction was not implementable as written.

**Acceptance:** `Pane` (and/or `sessions::Record`) carries the role it was spawned with, server-side; `parent_role_for` reads it; a SubOrchestrator pane can delegate Workers cross-dashboard; a forged request still cannot claim a role it wasn't spawned with. Regression tests for both directions.

## 2. Worker token budgets reset across supervised restarts

`--budget-tokens` measures spend against the **current child's own transcript**. A rot-, timeout-, or nudge-triggered restart mints a fresh transcript, so a run that restarts N times can spend up to N × budget rather than the ceiling the operator named. Documented in code comments at the seam.

**Acceptance:** the supervisor accumulates spend across restarts within one `zirv ctx agent` invocation and enforces the budget against the running total; a restart never resets the meter; test with a fake agent that restarts once and exceeds the budget only on cumulative spend.

Glubiz:
Fresh production evidence for the role-persistence gap (2026-08-28, zirv 2.32.0, macOS):

An **interactive orchestrator chat pane** (verb `chat`, `ZIRV_CTX_LAUNCH_MODE=interactive`, seat model fable, session `9e451bdd`) running inside a dashboard was refused delegation outright:

```
$ zirv agent codex "<brief>"
a worker may not delegate onward (delegation depth cap: 2)
[...]

PR #161, opened 2026-08-27
release: v2.31.0 — sandbox-escape allowlist, codex approval posture, and token-cost phases 1-4

Glubiz:
Combines six reviewed branches into a single release so the project ships **one tag** instead of six. Version 2.30.1 → **2.31.0**.

Supersedes #154, #153, #156, #157, #158, #159 — all now draft to prevent a separate merge, each closing when this lands. Every reviewed commit is carried over via `--no-ff` merges, so per-PR history and attribution are preserved.

## What ships

| From | Change |
|---|---|
| #154 | Sandbox-escape allowlist (closes #147), including a critical root-wide `find` bypass closed by lexical path normalization |
| #153 | Codex approval posture: config-override pinning, writable roots, network knob |
| #156 | Epic #155 phase 1: loud truncation, trimmed `common.md`, one deduped memory layer at the cacheable tail |
| #157 | Epic #155 phase 2: raw four-category token telemetry with delegation lineage |
| #158 | Epic #155 phase 3: canonical-content hash stamps and compose-time context dedupe |
| #159 | Epic #155 phase 4 (engine half): delta re-review and stop-on-no-new-findings |

## Merge integrity

Merged in dependency order: #154 → #153 → #156 → #157 → #158 → #159. Conflicts resolved preserving both sides in every case:

- **`permissions.rs`** — #154 added a `native_rules` parameter to `render_report()`; #153 added a codex-only caveat calling the old one-arg form. Resolved by printing the caveat, then calling the new two-arg form. Both behaviors kept.
- **`log.rs`** — #154's `SafetyDecisionRecord`/`read_safety_decisions` and #157's `tail_delegations` landed at the same insertion point. Independent functions, kept sequentially.
- **`Cargo.toml`/`Cargo.lock`** — per-branch bumps (2.31.0–2.36.0) collapsed to a single 2.31.0.
- **Vault pages** — two-sided entries for different phases, both kept.

**One semantic conflict `cargo build` did not catch.** #153 added a test constructing `CompileArgs{}` without the `escape` field #154 introduced. The two edits were on non-overlapping lines, so git reported no conflict and the library compiled — only the test build failed. Fixed by adding `escape: false`, matching every other such literal in the file. This is the class of defect a combined merge risks, and it is the reason the full test suite is the gate here rather than a build.

The flagged high-risk overlap — #157 and #158 are siblings on the phase-1 base and both edit `compile.rs` and `context_status.rs` — auto-merged without conflict, and both sides were verified present and coexisting afterward.

[...]

Issue #160, opened 2026-08-27
Approval-surface residuals from the #147 security review: restored-pane pin, relative-path find scans, commit-message misclassification


Glubiz:
Three residual findings surfaced during the #154 security review train (2026-08-27). None blocked that PR; all are worth tracking. Grouped because they all live in the same approval/safety surface.

## 1. Restored dashboard panes never receive the interactive launch-mode pin

`spawn_restored_pane` (`src/commands/ctx/dash/mod.rs`, ~line 3629) does not set `LAUNCH_MODE_ENV`. Only `fulfill_spawn_request`'s two callers and the orchestrator pane in `run_dashboard` do.

This is **not a regression** — restored panes were already headless-classified before #154 and they fail closed, which is the safe direction. But it means the bug #154 fixed for the orchestrator and spawn paths ("an operator's `defaultMode=auto` defeats interactive classification") still applies to a restored worker pane that a human is actually sitting in front of and watching in the dashboard.

**Acceptance:** a restored pane a human is watching classifies as interactive on the same terms as a freshly spawned one, with a regression test pinning it.

## 2. `is_root_wide_find_scan` still ignores relative paths

`find .. -iname id_rsa` escapes silently — confirmed by probe against `a6a0bb4`.

This is **pre-existing accepted scope**, not a miss in #154's three fix rounds: the original exact-match check never handled relative paths either, and both `is_root_wide_find_scan`'s own doc comment and `generated_path` elsewhere in `safety.rs` already document the non-goal — a text-only classifier cannot know whether the launch's working directory is itself at or above a sensitive root. It is also cwd-dependent, so unlike the `/`, `//`, `/.`, `~root` spellings that #154 closed, there is no clean environment-independent PoC.

**Acceptance:** decide explicitly whether to (a) keep the documented non-goal, (b) treat any relative start as unbounded-or-refuse, or (c) resolve against the known launch cwd where zirv has it. Record the decision; if (a), no code change — just make the doc comment state it as a deliberate ruling.

## 3. The safety classifier denies commit messages containing path-like strings

While committing the fix for finding #2 above, `git commit` with a heredoc message was **denied by zirv's own permission layer** because the message text contained `~/` and `/../`. The worker had to fall back to `git commit -F <file>`.

Path-like strings inside a commit message body are being classified as if they were command arguments. This is a real usability bug — and a self-inflicted one, since it fires exactly when committing work *about* path handling.

[...]

PR #171, opened 2026-08-27
release: v2.32.0 — sub-orchestrators, work groups, budgets, permits, capacity-aware gates (#155 phases 4-6)

### LINKED ISSUE #162: Heavy-worker slot leaks on a failed delegation and blocks all further zirv ctx agent runs
`zirv ctx agent` is refused by the heavy-worker gate while no heavy worker is running, and the leaked slot is not attributable to any session `zirv ctx status` lists.

## Observed, 2026-08-27 on macOS, zirv 2.31.0

At session start:

```
heavy workers: 0 of 1 slots in use
```

Two `zirv ctx agent claude` invocations were then made. The first failed immediately on a missing prompt file (empty stdin) and exited. The second was refused:

```
error: zirv ctx exec: refusing to start a new heavy worker -- 1 of 1 slots
already in use (supervise.max_heavy_workers).
```

`zirv ctx status` at that moment reported:

```
heavy workers: 1 of 1 slots in use

sessions:
  1d908765  claude  chat  pid 22906  1h   live
  6e5ac20f  claude  chat  pid 88105  22h  live
  d25b87d7  codex   dash  pid 676    22h  live
```

All three live sessions are `chat` or `dash`. None is a headless heavy worker. So the gate is holding a slot that corresponds to no listed session, and the operator has no way to identify or release it — the error message's own advice ("see `zirv ctx status` for the list of running sessions") points at a list that does not contain the offender.

## Why it matters

The slot appears to leak from a delegation that failed during startup: the run registers against the budget before it can fail, and nothing releases the registration on that path. Once leaked, `max_heavy_workers = 1` means **all** further delegation is refused for the lifetime of the state dir, with no visible cause and no documented way to clear it short of editing config or setting `ZIRV_CTX_SUPERVISE_MAX_HEAVY_WORKERS`.

This is self-inflicted denial of the delegation path, and it is invisible.

## Relationship to issue #155 Task 5.5

Task 5.5 ("heavy-**operation** permits replace workload-blind session counting") already plans to replace this accounting. Two things are worth carrying into that work rather than treating as separate:

1. **Registration must be released on every failure path**, including a delegation that dies before the worker starts. A permit acquired and then abandoned must not survive the process that acquired it.
2. **The refusal message must name what holds the permit** — session id and what it is doing — and whatever it names must actually appear in `zirv ctx status`. A capacity refusal an operator cannot attribute is worse than no gate.

There is a real irony worth recording: the delegation that was refused here *was the delegation implementing Task 5.5*. The bug blocks its own fix.

## Acceptance

- A delegation that fails during startup leaves the heavy-worker count exactly as it found it, with a regression test covering the early-failure path.
- A refusal names the holding session, and that session is listed by `zirv ctx status`.
- Some supported way exists to observe and release a stale permit without editing configuration.

Glubiz:
Single combined release for the remainder of epic #155 (phases 4-6) — one PR, one tag, version 2.31.0 → **2.32.0**. Same shape as #161. Supersedes #163, #164, #165 (set to draft; every commit is carried here via `--no-ff` merges, per-PR history preserved).

## What ships

| Task | Change |
|---|---|
| 4.1 | An active `zirv workflow` review gate is the single review source of truth — `HARNESS_PROMPT` (v10) + `ORCHESTRATOR_PROMPT` no longer stack three review rounds on one change |
| 5.1 | `PromptRole::SubOrchestrator` with a trimmed layer: delegation vocabulary, no HARNESS_PROMPT/roster, may never spawn another coordinator |
| 5.2 | Work groups: `zirv ctx group create|status|close`, persisted under `<state>/groups/` |
| 5.3 | Spawn-request lineage (`role`/`parent_session`/`work_group_id`) + depth cap (meta → sub-orchestrator → worker, never deeper) |
| 5.4 | Worker budgets `--budget-tokens`/`--max-tool-calls`: soft-warn at 80%, hard stop with `EXIT_BUDGET_EXHAUSTED` — checkpoints, never downgrades the model |
| 5.5 | Heavy-**operation** permits replace workload-blind session counting (RAII release on every path, dead-pid sweep) — structurally fixes #162 |
| 5.6 | `zirv ctx status` renders the work-group tree with per-child raw-class spend |
| 6.1 | `Capabilities::context_window_tokens` (claude: 200k / 1M for `[1m]`; codex: honest `None`) |
| 6.2 | Rotation gates become ratios of real capacity via pure `rot::token_gates`; explicit absolutes still win; None-capacity keeps today's exact thresholds |
| 6.3 | Quota pressure gates **new spawns** (warn 80% / refuse 95% / `--force` override), never rotation — a refusal costs nothing, a rotation destroys a warm cache |
| 6.4 | `docs/benchmarks/token-cost.md`: metric definitions, pinned measurement protocol, honest "not yet measured" rows |

## Verification

- Full suite on the combined tree: **2809/2809** under nextest AND serial `cargo test --test-threads=1`; build, fmt, clippy `-D warnings` all clean.
- Per-task presence checks after every merge stage (nothing dropped); `rot.rs` purity re-verified (zero fs/clock/env/net references).
- Two invisible merge seams (test-build breaks with no textual conflict) were caught and fixed: `CompileArgs.escape`-class and a stale `max_heavy_workers` assignment.
- **Independent review round: IN PROGRESS at PR-open time.** One sonnet reviewer over this full diff, with an explicit question on whether a forged file-drop spawn request with `force: true` can bypass the quota gate's hard refusal. Findings (or the clean verdict) will be posted as a comment before merge; treat that comment as part of this PR's record.

## Merge instructions

Merge with a **merge commit**, not squash (preserves the six task branches' reviewed history). The Protected Main ruleset needs 1 approval and GitHub forbids self-approval, so merge with `--admin` as for #161. CD then tags v2.32.0 and publishes.

After merge: close/delete the superseded branches (`feat/token-cost-p4-review-convergence`, `p5-sub-orchestrator`, `p5-work-groups`, `p5-heavy-permits`, `p5-lineage-budgets`, `p5-status-tree`, `p6-context-capability`, `p6-rotation-ratios`, `p6-spawn-gate`, `docs/token-cost-benchmarks`).

## Tracked follow-ups (not blockers)

- #169 — persist pane roles (sub-orchestrator chaining currently fails closed); accumulate budgets across restarts.
- #170 — the operator's target sub-orchestrator model (one layer, scope-bound work groups).
- #160 — security residuals from the #147 review. #162 closes with this merge.
- Benchmarks rows fill in once the new telemetry runs; epic #155's acceptance number still requires that measurement.
- 6.3 shadows dash's old 99% usage gate's refuse arm at default config — consolidation candidate.


Glubiz:
**Review status at session end:** the independent sonnet review of this full diff did NOT complete before the session closed — the reviewer agent dies with it. Before merging, a next session must run that review: full diff origin/main...release/2.32.0, priorities = (1) forged file-drop SpawnRequest with force:true bypassing pace::spawn_gate's hard refusal (trusted_interactive forgery discipline), (2) permit RAII/dead-pid-sweep correctness, (3) budget hard-stop races, (4) config trust directions for all new keys, (5) a third instance of the invisible merge-seam class. Post the verdict here; fix rounds update this branch in place. Recovery state: zirv memory key epic-155-endgame-plan.

Glubiz:
## Independent review round — verdict (completes the review promised in the PR description)

**Review record:** one Sonnet reviewer over the full `origin/main...release/2.32.0` diff with the five priorities from the session-end comment, plus a codex (gpt-5.6-terra) cross-review of the same diff. Their findings were triaged together, fixed in two rounds (16 commits, `161ed7c..739804d`), the fix range re-reviewed by a second independent Sonnet pass. The codex re-review of the fix range was skipped by operator instruction; codex's initial round stands as the cross-harness record.

### Priority verdicts (all five now closed)

- **P1 — forged file-drop `force`: CONFIRMED by both reviewers, fixed.** `req.force` from a file-dropped SpawnRequest was honored as the operator's own override. It can no longer lift the ≥95% hard refusal (`0aea04d`); operator `--force` on a trusted interactive invocation is unaffected. Worse than the asked question: unknown/forged `parent_session` resolved to Orchestrator (maximum privilege), letting a Worker pane self-escalate past the depth cap — unverified lineage now never grants coordinator spawns, while Worker spawns keep working for the legitimate dash-rejoin flow (`418564c`).
- **P2 — permit RAII/dead-pid correctness: multiple real defects, fixed.** Acquisition was count-then-create (TOCTOU) — now atomic `create_new` slot files with post-write claim verification and a grace-window sweep of crash-orphaned partial files (`94dd6d8`, `23a445d`). The permit recorded the script-runner's pid, not the heavy child's — `child_pid` now keeps the slot held while the real work runs (`b369627`). A `CtxConfig::load` error silently disabled permits entirely — now falls back to the default policy, fail-closed (`26db2f7`). The 600s wait-timeout escape hatch remains but warns loudly (`d8fffd1`).
- **P3 — budget hard-stop races: no tick race, but a real exit-path gap, fixed.** A worker writing an over-budget final transcript and exiting between polls returned its clean exit code — the final transcript is now evaluated on clean exits and returns `EXIT_BUDGET_EXHAUSTED` (failure codes are preserved) (`44f3735`). `--max-tool-calls` silently never counted for codex (its transcripts carry no tool-call events) — now refused up-front via a new `counts_tool_calls` adapter capability (`60f48a6`). Budgets still reset across restarts — already tracked in #169, unchanged.
- **P4 — config trust for new keys: one loosening found, fixed.** All new `score.*`/`pace.*`/`supervise.max_heavy_operations` keys are correctly `REPO_FORBIDDEN` with tests. But repo-layer `heavy_command_patterns` REPLACED the operator array (repo `[]` erased home patterns) — now lift-and-union, add-only, like `sandbox.extra_deny` (`161ed7c`).
- **P5 — third invisible merge seam: found, fixed.** `capabilities_for_model` was `#[allow(dead_code)]` with no production caller — every live scoring path passed `model: None`, so the `[1m]`/1M-context branch never fired and the epic's motivating bug (restarting a 1M seat at 16%) was not actually fixed. Now wired: a new `AgentAdapter::model_hint` resolves the session's real model from the transcript into `capabilities_for_model` on the live scoring paths; the incremental scorer persists the model across polls (checkpoint schema v1→2, old checkpoints rebuild cleanly) (`4e27fcd`). `rot.rs` purity re-verified untouched.

### Beyond the five priorities

- Work-group `child_limit`/`deadline_secs` were persisted but never read, and completed delegations logged `work_group_id: None` — `child_limit` is now enforced at both admission choke points, with admission rolled back on every post-admit failure path so failed spawns can't burn the limit; deadline renders a display-only OVERDUE marker; the group id is logged (`12dcab5`, `1ae6326`).
- The re-review of the fix range confirmed 10/11 fixes clean and surfaced the two round-2 findings above (admission rollback, permit slot hygiene); both fixed. No further confirmed findings — the loop terminated inside the 2-round limit.

### Accepted residuals (not blockers)

- Budgets reset across rot/nudge restarts, and sub-orchestrator pane roles are not persisted (honest chaining fails closed) — both tracked in #169.
- Raw-pid liveness has no start-time check (PID reuse can leak a slot, never bypass a limit) — pre-existing pattern shared with the session registry, predates this PR.
- Permits gate only script-runner commands by design; `completion_contract` is display-only (documented).

### Verification on the final tree

- Full nextest sweep: 2796 passed; serial `cargo test -- --test-threads=1` compatibility run: 2807 passed. Every failure name in both runs maps to this machine's documented environment baseline (broken-shell-PATH classes + wrap turn-signal); the two initially-outside-baseline names were run to ground: one confirmed platform-state-dir flake (passes 3/3 isolated), one pre-existing dishonest 2s test budget vs. real Windows `taskkill` cost (1.4–1.9s measured) — widened to 10s (`739804d`), not a code defect.
- `cargo build`, `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings` all clean. `wrap.rs`, `pace.rs`, and adapter argv untouched — no Linux verification round required.
- Obsidian vault updated for all behavior/contract changes (`535ff1e`).

**Verdict: merge-ready** once CI is green on `739804d`. Merge per the PR description: merge commit (not squash), `--admin`.

Issue #180, opened 2026-08-28
Flaky test: a_failed_exit_with_an_over_budget_transcript_keeps_its_failure_code races the live budget tick


Glubiz:
## Problem

`commands::ctx::exec::tests::a_failed_exit_with_an_over_budget_transcript_keeps_its_failure_code` fails ~3/20 runs (measured 2026-08-28 on macOS, identical rate on main and on release/2.33.0 — pre-existing, byte-identical test on both).

Root cause (from the 2.33.0 debug round): the test's `fail` mode writes an over-budget transcript (480_004 tokens vs a 10_000 budget) and exits 3, but `supervise_child` checks `try_wait` first and `on_tick` second — when a budget tick lands while the child is still alive with a partially written over-budget transcript, the supervisor kills it and `run_with` returns 77 (budget exhaustion) instead of the child's own 3. Captured: `budget exhausted (120012/10000)` after 6 of 24 assistant events.

Both outcomes are legitimate production behaviour; the race is in the test's premise (transcript must stay under budget for the child's whole lifetime and go over only at exit). An atomic staging+mv in the fixture shrinks the window ~1000x but cannot close it.

## Acceptance
- [ ] Decide the test's premise: either stage the transcript atomically at exit (accepting the tiny residual window), or split the test into the two deterministic halves (child-exit-first vs tick-first) and assert each explicitly.
- [ ] 0/40 failures in isolated serial loops on the machine that reproduced it.

Glubiz:
**Status 2026-08-28** (audit of `release/2.33.0` @ 016495a)

**Not fixed yet.** f5def66 fixed a *different* restart-accumulation test; the fixture for this one (`tests/fixtures/fake-agent.sh`) still incrementally writes an over-budget transcript before exiting 3, so the live budget-tick race remains.

**Missing:** a deterministic ordering between the fixture's transcript writes/exit and the supervisor's budget tick.

Glubiz:
Duplicate of #203, which pins the same root cause with the actionable fix location (exec.rs budget HardStop path returns immediately instead of draining to EOF the way the limit_hit path does). Closing in favor of #203.

Issue #152, opened 2026-08-26
sessions: pid liveness has no start-time disambiguator — recycled foreign-uid pid keeps a stale record alive


Glubiz:
Follow-up from PR #151 review (issue #146 fix).

`sessions::is_alive` now correctly treats `kill(pid,0)` EPERM as alive so sandboxed pane children can see live sessions. Trade-off: when a registered session's pid exits and the OS recycles that pid for an unrelated process under another uid, `kill` returns EPERM forever, so the stale record is never swept — it shows as Live in status/dashboards, `short_is_live` blocks restores, and dash token dirs can leak.

`Record` stores only a bare pid. Acceptance:
- Store a process start-time (or equivalent disambiguator) at registration and compare it during liveness checks (macOS: proc_pidinfo/sysctl kern.proc.pid; Linux: /proc/<pid>/stat field 22; Windows: existing mechanism unaffected).
- A recycled pid then reads as dead and the record is swept as before.
- Unit tests through the existing seams; missing/unreadable start-time degrades to today's behavior (EPERM = alive), never to false-dead.

Glubiz:
**Status 2026-08-28** (audit of `release/2.33.0` @ 016495a — cited commits are on the pending release branch, not yet on `main`)

**Partially implemented.** 98e53dd added a recycled-PID guard, but only on the `ctx kill` path.

**Missing:** start-time disambiguation in the *shared* pid-liveness check (`src/commands/ctx/sessions.rs`), which sweeps and restores still use — a recycled foreign-uid pid can still keep a stale record alive everywhere except `ctx kill`.

PR #182, opened 2026-08-28
fix(ctx): approval-surface residuals from the #147 security review (#160)

### LINKED ISSUE #160: Approval-surface residuals from the #147 security review: restored-pane pin, relative-path find scans, commit-message misclassification

Glubiz:
Closes #160.

## What

All three residuals from the #154 security review train:

1. **Restored dashboard panes now restore their original launch mode.** `Pane` carries a `launch_mode` field derived at spawn time from the actual `turn_env` handed to the child (so it can never drift from what the child inherits); `on_quit` rosters it as `RosterPane::interactive` (`#[serde(default)]` → `false`, fail-closed for pre-upgrade roster files); `spawn_restored_pane` restores the pane with the launch mode it originally had. A restored orchestrator pane a human is watching classifies interactive again; a worker pane spawned headless stays headless across quit+restore — no escalation.
2. **The relative-path scope of `is_root_wide_find_scan` is recorded as a deliberate ruling** (issue #160 finding 2, option (a)): doc comments on `is_root_wide_or_whole_home_path`/`is_root_wide_find_scan` state the non-goal explicitly, and a pinning test documents that `find .. -iname id_rsa` is deliberately outside the root-wide rule's scope.
3. **Regression tests for opaque commit messages**: a `git commit -m` (and heredoc-built) message containing `~/`, `/../`, and a redirection-looking character is not classified on its message text (pins the PR #140 `redact_opaque_message` fix per the issue's acceptance criteria).

Structural hardening from the review round: `build_turn_env` now takes a mandatory `LaunchMode` parameter and pushes the `ZIRV_CTX_LAUNCH_MODE` pin itself (both success and adapter-resolution-failure branches), so a future spawn path that forgets the pin is a compile error instead of a silent headless classification — this PR was the third occurrence of that bug class.

Known residual (documented in-code, out of scope): `handover::build_turn_env` does not push the pin, so `Pane::launch_mode()` can go stale across a handover.

## Verification

- Windows: `cargo build`, `cargo nextest run --no-fail-fast`, serial `cargo test`, `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings` — all green; failing-name diff vs the machine baseline is empty (only the known #180 flake and the documented `wrap::win` baseline failure).
- Linux (`rust:1-bookworm`, non-root, `git -c core.autocrlf=false archive` export): `cargo clippy --all-targets -- -D warnings` clean; `cargo test --bin zirv dash::` 368/368.
- Review: medium-effort review round (3 confirmed findings, all fixed — the unconditional-interactive-pin escalation, the pin-at-call-sites bug class, missing env-pair coverage) plus a clean delta re-review.

https://claude.ai/code/session_017QEYoGbDzvGYAT2LJgUZBQ

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2693,7 +2693,7 @@ checksum = "b97154e67e32c85465826e8bcc1c59429aaaf107c1e4a9e53c8d8ccd5eff88d0"
 
 [[package]]
 name = "zirv"
-version = "2.33.0"
+version = "2.33.1"
 dependencies = [
  "clap",
  "console",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zirv"
-version = "2.33.0"
+version = "2.33.1"
 edition = "2024"
 authors = ["Jonathan Solskov <josj@zirv.io>"]
 description = "A CLI tool for executing developer-defined YAML scripts with support for interactive commands and OS-specific execution."
--- a/src/commands/ctx/dash/mod.rs
+++ b/src/commands/ctx/dash/mod.rs
@@ -4014,6 +4014,63 @@ fn partition_restore_selection(
     (to_spawn, deferred)
 }
 
+/// Builds the `turn_env` a restored dashboard pane spawns with -- everything
+/// `spawn_restored_pane` pushes ahead of `Pane::spawn`: the base env `build_
+/// turn_env` produces, a fresh spawn-request channel of its own (Security
+/// review Finding 1), the roster's group binding when the candidate carried
+/// one (Security review Finding 6), and the durable interactive-launch pin
+/// (issue #160 finding 1). Factored out of `spawn_restored_pane` so the exact
+/// fields it adds are pinned directly, independent of a real pty spawn's
+/// behavior -- the same reasoning `spawn_launch_mode_pin`'s own doc comment
+/// gives for testing that pin as a pure function rather than reading a real
+/// spawned child's own environment back.
+///
+/// Returns the built `turn_env` alongside the freshly minted pane channel
+/// path: `spawn_restored_pane` needs both, the env to spawn with and the
+/// path to hand the spawned `Pane` via `set_intake_dir`.
+fn restored_pane_turn_env(
+    cfg: &CtxConfig,
+    state: &StateDir,
+    repo: &Path,
+    candidate: &roster::RosterPane,
+    requests_dir: &Path,
+    errors: &mut Vec<String>,
+) -> (Vec<(String, String)>, PathBuf) {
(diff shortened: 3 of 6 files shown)


Glubiz:
Superseded: per the operator, #152/#155/#160 plus the CI/CD split ship as one combined PR from release/2.34.0. This branch is merged into that release branch unchanged.

PR #183, opened 2026-08-28
fix(ctx-sessions): start-time disambiguator sweeps recycled-pid records (#152)

### LINKED ISSUE #152: sessions: pid liveness has no start-time disambiguator — recycled foreign-uid pid keeps a stale record alive

Glubiz:
Closes #152.

## What

`sessions::Record` gains `start_time: Option<u64>` (`#[serde(default)]`, epoch seconds) as the pid-recycling disambiguator the #146 EPERM-is-alive fix deliberately deferred:

- Stamped at registration from `process_start_secs(pid)` — derived as `now - process_age_secs(pid)` from the existing `ps -o etime=` reader; no new platform APIs. Re-stamped at both places `record.pid` is repointed at a spawned child (`SessionGuard::adopt_child_pid`, dash `Pane::spawn`), so the recorded start time always describes the process the pid names.
- New `record_is_alive(&Record)` drives `short_is_live` and the `sessions::list` sweep: unix `kill(pid,0)` success → alive unconditionally; `ESRCH` → dead; `EPERM` → alive unless a freshly derived start time mismatches the recorded one beyond `START_TIME_TOLERANCE_SECS` (300 s) — then the record reads dead and is swept. A recycled foreign-uid pid no longer keeps a stale record Live forever.
- Degrade rule (the forbidden-regression guard): a missing or unreadable start time on either side always falls back to today's EPERM=alive behavior — never false-dead. Old records without the field deserialize to `None` and behave exactly as before.
- Windows liveness is deliberately unchanged; bare `is_alive` remains for the non-record callers (`permit`, `supervise`, dashboard owner file), and `permit.rs`'s doc comment now records that `PermitRecord` deliberately keeps the bare-pid probe. The 300 s tolerance and the kill-path's `RECYCLED_PID_TOLERANCE_SECS` (5 s) are cross-referenced as serving different mechanisms.

The 300 s tolerance is sized to the failure mode: recycled-pid staleness is an hours-to-days phenomenon, while NTP steps, `ps` etime rounding, and derivation slack are seconds-to-minutes noise that must never sweep a live session.

## Verification

- Windows: all five gates green; failing-name diff vs the machine baseline is empty (only the known #180 flake and the documented `wrap::win` baseline failure).
- Linux (`rust:1-bookworm`, non-root): `cargo test --bin zirv sessions::` 64/64 (including the new `cfg(unix)` EPERM/pid-1, adopt-repoint, and old-format regression tests); `cargo clippy --all-targets -- -D warnings` clean.
- Review: medium-effort review round (6 findings — two confirmed false-dead bugs (supervisor-vs-child start_time baseline, wall-clock-step sweep), permit doc drift, a test flake, two robustness items — all fixed) plus a clean delta re-review.

https://claude.ai/code/session_017QEYoGbDzvGYAT2LJgUZBQ

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2693,7 +2693,7 @@ checksum = "b97154e67e32c85465826e8bcc1c59429aaaf107c1e4a9e53c8d8ccd5eff88d0"
 
 [[package]]
 name = "zirv"
-version = "2.33.0"
+version = "2.33.2"
 dependencies = [
  "clap",
  "console",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zirv"
-version = "2.33.0"
+version = "2.33.2"
 edition = "2024"
 authors = ["Jonathan Solskov <josj@zirv.io>"]
 description = "A CLI tool for executing developer-defined YAML scripts with support for interactive commands and OS-specific execution."
--- a/src/commands/ctx/dash/mod.rs
+++ b/src/commands/ctx/dash/mod.rs
@@ -7678,6 +7678,7 @@ mod tests {
             owner_pid,
             safety_policy_sha256: None,
             role: None,
+            start_time: None,
         }
     }
 
--- a/src/commands/ctx/dash/pane.rs
+++ b/src/commands/ctx/dash/pane.rs
@@ -954,8 +954,17 @@ impl Pane {
         // from another's. Stamp the child's real pid instead. `process_id`
         // returns `None` on a platform that cannot report it; there we leave
         // the dashboard's pid rather than a bogus one.
+        //
+        // Review round 2 finding 1 (issue #152): `start_time` must move with
+        // `pid` in the same breath, via the same `sessions::process_start_secs`
+        // reader `Record::new` itself used -- see `SessionGuard::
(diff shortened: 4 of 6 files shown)


Glubiz:
Superseded: per the operator, #152/#155/#160 plus the CI/CD split ship as one combined PR from release/2.34.0. This branch is merged into that release branch unchanged.

Issue #303, opened 2026-09-02
Headless Compact tier for codex: implement headless_resume_cmd via codex exec resume (#283 follow-up)


Glubiz:
Deferred from the release/3.9.0 review round (2026-09-02).

**Gap.** #283 lets headless exec/loop act on the Compact rot tier by compacting in place and resuming the same session. Only the Claude adapter implements headless_resume_cmd; codex has no override, so after the review fix a codex Compact verdict maps straight to the existing restart path (no cooldown wasted, but no in-place compaction either). Codex sessions therefore still pay a full restart plus handoff re-priming for every Compact verdict.

**Evidence.** The installed codex-cli 0.152.0 documents `codex exec resume [SESSION_ID] [PROMPT]` with PROMPT set to - for stdin, which is exactly the shape headless_resume_launch needs.

**Proposal.** Implement CodexAdapter::headless_resume_cmd (and the matching compact_command if still None) on top of codex exec resume, flip supports_headless_compact to true for codex, and reuse the shared compaction-verification path (transcript check, cooldown, fail-closed fallback to restart) unchanged.

**Acceptance.** A codex headless loop hitting Compact compacts in place and continues the same session when the transcript confirms compaction; an unverified compaction still falls back to restart; the #283 codex fallback test is updated to the new contract; all five gates pass.

Glubiz:
Finding while implementing (2026-09-09, codex-cli 0.153.4): `codex exec resume [SESSION_ID] [PROMPT]` is real (`-` reads the prompt from stdin), but codex documents NO compaction directive anywhere -- no subcommand, flag or config key across `codex --help`, `codex exec --help` and `codex exec resume --help`, and the interactive slash-command set cannot be probed non-interactively. So `compact_command()` stays `None` and `supports_headless_compact()` stays `false` for codex; the Compact tier still maps to restart-with-handoff, because there is nothing verified to trigger in place.

What the 3.38.0 release does ship from this issue: `CodexAdapter::headless_resume_cmd` on `codex exec resume`, plus a new `AgentAdapter::resume_target` hook that recovers codex's own minted session id from the rollout's `session_meta.payload.id` (zirv's session id is not codex's) and fails closed to the one-attempt path when it cannot. Net effect: the #318 bounded contract retry now covers codex workers.

Suggest narrowing this issue to: 'headless Compact for codex once codex-cli exposes a compaction directive' (a codex-cli feature request), or closing it with that note.

Glubiz:
[...]

Issue #295, opened 2026-09-02
Memory bank round 3: reversible journaled writes with conflict check, and a session-scoped tier with explicit promotion


Glubiz:
## Origin

Prime Agent analysis, tracked in #279. Prime's `/refine` stores a full `before`/`after` snapshot per applied edit (`packages/coding-agent/src/core/refinement/refinement.ts:788`), so `rollbackProposal` can build inverse edits in reverse order (`refinement.ts:813-845`), and the apply phase re-reads state immediately before writing and rejects any edit whose target moved since planning (`refinement.ts:735-749`) — verified, with 1516 lines of tests covering baseline-conflict rejection, atomic save, and cross-session rollback through a global `refinements.jsonl`. Its second good idea is scope: the default store is **session-local**, `--global` is opt-in and policy-restricted to cross-session lessons (`refinement.ts:140-149`) — the cheapest known defence against poisoning a durable bank with one-off session state. The anti-pattern to cite and avoid: the Python kernel writes the same durable store directly through `rlm.harness.upsert`/`create_*`/`delete` (`prime-agent-runtime/src/rlm/harness.py:303-676`) with no review, no snapshot, no history record and no base-prompt guard, so "small, evidence-backed updates" is unenforced and those writes are not rollback-able. zirv's one-validated-CLI-entry-point posture is right; keep it.

## Problem in zirv today

1. **No versioning, no rollback, no conflict detection.** `memory::remember` (`src/commands/ctx/memory.rs:1287-1343`) lists, deletes every file with a matching key (`:1296-1300`), then writes. The old body is gone: nothing on disk can restore it.
2. **Last-writer-wins is explicit and documented.** Two concurrent `remember`s on one key each miss the other's file, both write, and the loser is deleted by the `(written, path)` maximum rule (`memory.rs:1315-1338`). Correct as a convergence rule, wrong as a contract for parallel worktree sessions, which this repo runs routinely.
3. **`forget`/`verify` are equally unrecoverable** (`memory.rs:1356`, `:1379`, and the scoped variants at `:1024`, `:1112`).
4. **Only two scopes exist**: `MemoryScope { Shared, Private }` (`memory.rs:282-285`). Every `remember` is at least repo-durable under `<state>/memory/<repo_slug>/` (`state.rs:308`) and immediately joins the injected memory layer through `retrieval.rs`. There is no place to put "true for this session only".

### Journaled, reversible writes

[...]

Issue #294, opened 2026-09-02
zirv ctx measure: transcript-derived proportionality metrics with committed baselines


Glubiz:
## Origin

Prime Agent analysis, tracked in #279. Prime keeps five offline JSONL analysers over its session store (`scripts/read-tool-stats.mjs`, `scripts/edit-tool-stats.mjs`, `scripts/session-context-stats.mjs`, `scripts/tool-stats.ts`, `scripts/cost.ts`) that answer "did a harness change help" from transcripts alone. The sharp parts, all verified by reading the scripts: partial-vs-full read rate reported as a **median per session** rather than a pooled rate, so one heavy session cannot dominate; an edit **inflation ratio** computed by stripping the shared prefix/suffix of `oldText`/`newText` to get core changed bytes, then `totalEditBytes / coreBytes` (`scripts/edit-tool-stats.mjs:165-192`); context utilisation **before the first compaction** as well as at peak; and `--since` defaulting to the birth time of the tool's own implementation file (`scripts/read-tool-stats.mjs:129-137`), so a run measures only sessions produced by the current version of the thing being tuned. What Prime gets wrong and zirv should not copy: none of it is wired, no thresholds, no committed baseline, no tests — a regression is only caught if someone remembers to run the script.

## Problem in zirv today

zirv measures its own artefacts, never agent behaviour.

1. `zirv ctx compile --measure` (`src/commands/ctx/compile.rs:896-1012`) renders a per-layer byte/token table for the prompt prefix — what zirv *sends*, not what the session then *does*.
2. `zirv workflow stats` (`src/commands/workflow/telemetry.rs:882`) aggregates locally recorded workflow events into phase stats and cache-hit ratios (`telemetry.rs:238`) — workflow-scoped, and only for sessions that ran a workflow.
3. `zirv ctx status`/`zirv ctx usage` report the four raw token classes and rate-limit window percentages (`usage.rs`, `status.rs`).

Nothing answers: how often did a worker read a whole file it needed thirty lines of; how much of an edit payload was unchanged context the model re-transmitted; how full did the context get before the first compaction; how many turns per user message. The material is already parsed — both adapters implement `transcript_path` (`adapters/mod.rs:1195`) and `window::sum_transcripts` (`window.rs:444`) already walks transcript trees — but no verb asks these questions, and there is no stored baseline to diff a prompt-layer change against.

### The verb

[...]

Issue #272, opened 2026-09-01
screen.rs round 2: truncation finding, role-marker/invisible-unicode/opaque-run markers, source-aware action matrix, memory write-cadence signal


Glubiz:
## Origin

Ruflo round-2 analysis, tracked in #278. Source: `v3/@claude-flow/hooks/src/workers/channel-guard-worker.ts` (mid-message `system:`/`assistant:` role markers at offset > 0, ≥80-char base64/hex runs, zero-width and bidi control characters; wired into inter-agent `sendMessage`), `v3/@claude-flow/security/src/tool-output-guardrail.ts` (**truncation is itself a finding**: the unscanned tail is reported as a medium finding instead of implicitly clean; a severity → action table), `v3/@claude-flow/plugin-agent-federation` PII pipeline (a **confidence × trust-level → action matrix** on the outbound path, 44 tests), `hooks/src/workers/memory-poison-forensics.ts` (z-score on write cadence and size per writer). Ruflo's guardrail suite has zero false-positive tests; keep zirv's corpus discipline.

## Problem in zirv today

`screen.rs` (#243) is pure and screens mail, repo-provided context, shared memory headers and transcript tails for `PromptInjectionMarker` (phrase list), `CredentialShape` and `HighEntropyRun`, with a posture of label + cap + flag. Gaps found by comparing rule sets:

1. **Truncation is silent.** `IncrementalScorer::poll` screens newly appended bytes or a 64 KB tail; the unscanned remainder is implicitly clean and the report does not say a cap was hit.
2. **Missing cheap, high-precision rules:** a role label (`system:`, `assistant:`, `user:`, `developer:`, `<system>`, `[INST]`) appearing mid-text (byte offset > 0, start of line) is a strong signal for text that will be concatenated into a prompt and is not in the phrase list; Unicode `Cf` (zero-width) and bidi controls (`U+202A..U+202E`, `U+2066..U+2069`) are invisible carriers; long base64/hex runs are already partly covered by `HighEntropyRun` but not length-typed.
3. **One severity table for all sources.** Repo-owned text, peer-session mail and operator input are screened identically. Ruflo's matrix makes repo text strictly harsher than operator text at the same confidence; zirv already knows the source (`PromptSource`, `mail.from_session`, `Entry.source`).
4. **No write-cadence signal on the shared memory bank.** A compromised or looping worker flooding `.zirv/memory/` is only caught by size caps.

### 1. `ScanTruncated` label

[...]

PR #323, opened 2026-09-03
release: 3.12.0 - close #267 #275 #264 #293 #299 #306 (harness batch 2)

### LINKED ISSUE #264: Cost ledger: price delegations, populate spawn-tree attribution, zirv ctx spend, dashboard aggregate row
## Origin

Ruflo round-2 analysis, tracked in #278. Source: `plugins/ruflo-cost-tracker/scripts/track.mjs` + `_prices.mjs` (a `Stop` hook parses `~/.claude/projects/<enc-cwd>/*.jsonl` `message.usage` incl. cache read/write and prices it from a hard-coded table), `plugins/ruflo-cost-tracker/REFERENCE.md` (a "By agent: coder $5.20" report that **no script produces**), ADR-147 P2 (spawn tree with per-tree cost attribution, planned, not built), ADR-379 (statusline `session 0% · week 51% · $49.48` segments). Ruflo claims per-agent cost attribution and cannot deliver it because it does not own the pane roster. zirv does.

## Problem in zirv today

- `log::Delegation` already records the four raw token classes, `wall_ms`, `exit_code`, `outcome`, `agent`, `model` per delegation, and `status::group_tree_lines` reads the rows back for the group tree. Nothing **aggregates** them: there is no per-session, per-group, per-harness, per-model, or per-day spend view, and no dollar figure anywhere in zirv (config is tokens and percentages only).
- `work_group_id` on delegation rows is always `None` in practice; `TelemetryEvent.parent_session_id` exists and is never populated. Attribution up the spawn tree is therefore impossible today.
- `workflow/telemetry.rs` records token classes and `duration_ms` per event but no cost, and `zirv workflow stats` cannot say what a workflow cost.
- The dashboard shows per-pane state only; no aggregate row (workers running, spend this window, failures).

The question the `Delegation` doc-comment says it exists to answer, "was delegating cheaper than doing it on the orchestrator seat", cannot be answered without this.

### 1. Price table with an `as_of` date

`src/commands/ctx/price.rs`, pure: `fn price(model: &str, usage: &TranscriptUsage, table: &PriceTable) -> Option<Micros>`. Built-in table for the models zirv already names (`handover::equivalent_model` list) with `as_of: "2026-09-01"`; operator override `~/.zirv/prices.toml` (REPO_FORBIDDEN as a concept: a repo may never set prices). Currency as integer micro-USD (ADR-322C's rule: never floats in ledgers). A table older than `price.stale_after_days` (default 90) makes every cost line render `~$x (prices as of …)` rather than silently wrong.

### 2. Attribution fields written at the source

- `agent.rs` / `dash/spawnreq.rs`: populate `work_group_id` and `parent_session` on every delegation row (the parent is known at spawn time; today it is dropped on the pane path).
- `workflow/telemetry.rs`: populate `parent_session_id`; add `cost_micros: Option<u64>` and `price_as_of`.
- New optional `task_class` on `Delegation` (`review | test | implement | research | other`, derived from the workflow step that spawned it, or `--task-class` flag) so later analysis and routing can group outcomes by kind of work.

### 3. `zirv ctx spend` (new verb, read-only)

```
zirv ctx spend [--session <short>|--group <id>|--since 24h|--by harness|model|task-class|worker] [--json]
```
Reads `delegations.jsonl` plus the session's own transcript usage (already computed by `exec.rs::TranscriptUsage` / the usage tee) and prints a table: rows by the chosen dimension, columns `runs · ok · failed · input · cache-read · cache-write · output · wall · cost`. Totals line. `--json` for scripts. Deterministic ordering (by cost desc, then name).

### 4. Surfaces

- `zirv ctx status` gains one line: `spend: $1.84 this session · $12.10 this 5h window (prices as of 2026-09-01)`.
- Dashboard: an aggregate row above the roster: `workers 3 running · 1 failed · $4.20 · five_hour 41%`. Every cell follows a **source contract**: `Option<(value, Source, age)>`; a cell without a live source renders `--`, never a default number, and a test asserts no literal metric appears in a render path (Ruflo's statusline renders `patternsLearned: 156` as a hard-coded literal, `statusline/index.ts:517`).
- `zirv workflow stats` adds cost per workflow and per step.

### 5. `zirv report` friendly

`zirv ctx spend --json` output is the input the routing-feedback issue consumes; keep the schema versioned (`"schema": 1`).

## Files to touch

- new `src/commands/ctx/price.rs`, `src/commands/ctx/spend.rs`
- `src/commands/ctx/mod.rs` (verb), `status.rs`, `log.rs` (`task_class`, owned row parsing already exists), `agent.rs`, `dash/spawnreq.rs`, `dash/` render + roster
- `src/commands/workflow/telemetry.rs`, `stats` rendering
- `src/commands/ctx/config.rs` (`price.*` keys)
- docs: `Modules/Usage and Pacing.md`, `Modules/Ctx Subsystem.md`, dashboard page

## Acceptance criteria

- [ ] `price()` is pure and table-tested for each token class incl. cache read/write; unknown model → `None`, never 0.
- [ ] Every new delegation row has `parent_session`, `work_group_id` (when in a group) and `task_class` populated; a test spawns via the pane path and asserts the row.
- [ ] `zirv ctx spend --by harness` over a fixture `delegations.jsonl` prints deterministic totals matching a hand-computed expectation.
- [ ] Dashboard aggregate row renders `--` when no delegation has happened, and a render-path test fails if any metric literal is hard-coded.
- [ ] Stale price table (as_of older than threshold) is flagged in every cost line.
- [ ] `zirv ctx status --brief` unchanged in byte size except the one spend line (the token-cost baseline in `docs/benchmarks/token-cost.md` is re-measured).

## Tests

Pure price tests; spend aggregation over fixtures (deterministic order); attribution round-trip; render contract test; clippy/fmt; full five gates before PR.

## Non-goals

- Budget enforcement in dollars (pacing stays on usage windows and token budgets; this issue only makes cost visible).
- Pricing the orchestrator seat's own turns beyond what the usage tee already provides.

## Size / dependencies

M. Independent; the routing-feedback issue depends on `task_class` from this one.

### LINKED ISSUE #267: Writer vs reader permits and one-writer-per-tree for delegated workers
## Origin

Ruflo round-2 analysis, tracked in #278. Source: `v3/@claude-flow/codex/src/dual-mode/orchestrator.ts` `partitionLevel` (a `maxConcurrent` cap **and a separate `maxWriters` cap**; writers must hold distinct worktrees; dependency levels with cycle detection; `dependencyFailure: cancel|skip`), and Ruflo's CLAUDE.md "Concurrent Automated Development" rules: *never two writers in one worktree; read-only agents may share a checkout; only the integration owner edits shared manifests and lockfiles.* This part of Ruflo is real, tested and CI-gated (`scripts/audit-codex-integration.mjs`).

## Problem in zirv today

`permit.rs` has one pool: `supervise.max_heavy_operations` (default 1) slots, taken by `BUILTIN_HEAVY_PATTERNS` (cargo build/test/nextest/clippy/…) and `heavy_command_patterns`. Two consequences:

- A review worker and a research worker (both read-only, both cheap on the tree) queue behind the same single slot as an implementer running `cargo test`. Memory from earlier PRs says the practical rule became "limit concurrent heavy workers to two", enforced by hand.
- Nothing stops two *writing* workers from being dispatched into the same checkout. The worktree skill isolates a *workflow*, not a *worker*; `--workdir` validates a directory, it does not allocate one. Two implementers in one tree is the classic corrupted-diff failure.

### 1. Declared worker mode

`zirv agent <name> "<prompt>" --mode read-only|writing` (default `writing`, because a wrong "read-only" is worse than a wrong "writing"). The workflow engine sets it automatically for auto-spawned workers: `review run` and `test changed`/`verify` are `read-only`; anything spawned from the implement step is `writing`. The mode is written into the `SpawnRequest`, the `WorkerEnvelope` (delegation-envelope issue: `read-only` implies `tools ∌ edit`, `paths = []`), and the `Delegation` row.

### 2. Two permit pools

`permit.rs` gains a second pool `supervise.max_writers` (default 1) alongside `max_heavy_operations` (which becomes the pool for heavy *commands*, unchanged semantics). Rules:
- A `writing` worker needs a writer permit for its whole lifetime **and** a heavy permit while it runs a heavy command (as today).
- A `read-only` worker never takes a writer permit and takes a heavy permit only for heavy commands.
- Same `create_new` contention files (`<state>/permits/writer-<n>.json`) and the same dead-owner sweep.

### 3. One writer per tree

A writer permit records the canonical `cwd` (or worktree path). Dispatching a second `writing` worker into a tree that already has a live writer is refused with `AutoSpawnSkipped { Reason: WriterBusy }`-style announcement, unless `--worktree` is given, in which case zirv allocates `<repo>/.zirv/worktrees/<short>` via `git worktree add` from the session's base and passes it as `--workdir`. Read-only workers may share any tree.

### 4. Integration-owner rule

The orchestrator prompt (`claude::ORCHESTRATOR_PROMPT`, codex equivalent) gets one sentence: shared manifests and lockfiles (`Cargo.toml`, `Cargo.lock`, `package.json`, lockfiles) are edited only by the orchestrator or one designated integrator; writers touching them must say so in their report. Prompt-only, since zirv cannot see file ownership per worker without the envelope's `paths`.

## Files to touch

- `src/commands/ctx/permit.rs` (second pool, tree key), `agent.rs` (`--mode`, `--worktree`), `dash/spawnreq.rs` (`mode`, `worktree` in request/ack), `log.rs` (`mode` on `Delegation`), `config.rs` (`supervise.max_writers`, REPO_FORBIDDEN like `max_heavy_operations`)
- `src/commands/workflow/engine.rs` (`auto_spawn_decision` sets mode)
- `src/commands/ctx/adapters/claude.rs`, `codex.rs` (one sentence each)
- docs: `Modules/Ctx Supervisors.md`, `Modules/Usage and Pacing.md`

## Acceptance criteria

- [ ] Two `read-only` workers run concurrently with `max_writers = 1`; a second `writing` worker in the same tree is refused with a one-line reason and `retryable: true`.
- [ ] `--worktree` allocates a fresh worktree from the session base, and the writer permit records it; the worktree is listed by `zirv ctx status` and reclaimed by the existing sweep when the worker exits.
- [ ] `auto_spawn_decision` truth table extended: review/test/verify spawns are `read-only`.
- [ ] Permit files for writers survive a crash and are swept exactly like heavy permits (dead-owner test).
- [ ] `delegations.jsonl` rows carry `mode`.

## Tests

`permit.rs` pool tests (contention, sweep, tree key, case-fold on Windows); `agent.rs` flag tests; engine truth table; clippy/fmt; full five gates before PR.

## Non-goals

- Path-scoped leases between concurrent writers (Ruflo's `InMemoryFencedLeaseReference`). With one writer per tree they are redundant; revisit only for shared resources such as a release branch.
- Cross-harness enforcement inside Codex's own sandbox.

## Size / dependencies

S–M. Pairs with the delegation-envelope issue (`read-only` maps onto the envelope) but does not depend on it.

### LINKED ISSUE #275: zirv context lint: budget headroom, cross-layer duplicate and contradiction detection, proportionality metrics
## Origin

Ruflo round-2 analysis, tracked in #278. Source: `v3/@claude-flow/guidance/src/analyzer.ts` (a CLAUDE.md analyzer scoring 6 weighted dimensions incl. *enforceability* and *compilability*, with **mechanical size-budget rewriters** `optimizeForSize` at compact 80 / standard 200 / full 500 lines that split oversized sections, trim code blocks and dedupe rules) and `compiler.ts` (rule extraction by ID, `must/never/always/require` families → implicit rules, local overrides root by ID). The weights are hand-tuned and unvalidated; the byte budgets and the duplicate-rule detection are the useful part. Ruflo's ADR-172 also measured that a CLAUDE.md-heavy project directory costs **$1.56 per headless call** (56k cache-creation tokens) versus $0.34 from a clean directory: injected text is the dominant cost of every worker.

## Problem in zirv today

Round 2 of the wrapper redesign (PR #261) had to *manually* find that the fork/model policy was stated four times across `DEFAULT_PROMPT`, `claude::ORCHESTRATOR_PROMPT`, `.zirv/context/claude.md` and `CLAUDE.md`, that codex.md contradicted the codex orchestrator prompt, and that `common.md` sat at 4047 of 4096 bytes. zirv has the budgets (`context.max_*_bytes`, `prompt.max_repo_bytes`) and truncates on overflow with a `context-truncated` log line, and `zirv ctx compile --measure` prints per-layer bytes, but there is no check that:

- a rule sentence appears in more than one layer (duplication = wasted prefix bytes on every turn and a contradiction risk when one copy drifts);
- two layers give conflicting instructions on the same subject (the codex.md "NO subagents" case);
- a layer is within N bytes of its budget (headroom warning before the next edit silently truncates);
- the injected text is proportional (share of imperative sentences vs narrative), which is the metric the whole proportionality redesign was about.

### `zirv context lint [--json] [--budget compact|standard|full]`

Pure analysis over the same inputs `compile.rs` assembles (built-in prompts, harness prompts, role files, `.zirv/context/*.md`, generated `CLAUDE.md`/`AGENTS.md`), producing findings with stable ids:

1. **Budget headroom** (`CTX001`): per layer `bytes / budget`, warn at ≥ 90 %, error at overflow (which today is only logged at compile time).
2. **Duplicate rule** (`CTX002`): sentence-level near-duplicates across layers. Normalise (lowercase, strip punctuation/backticks), split on sentence boundaries, keep sentences containing an imperative marker (`must|never|always|only|do not|before|after|every`), compare with token-Jaccard ≥ 0.6. Report both locations and the byte cost of the second copy.
3. **Contradiction candidate** (`CTX003`): two imperative sentences sharing ≥ 3 content tokens where one contains a negation marker and the other does not (e.g. "NO subagents" vs "delegate to subagents"). Heuristic, reported as `candidate`, never blocking.
4. **Proportionality** (`CTX004`, info): per layer, count of imperative sentences, narrative sentences, code fences, and bytes; a one-line ratio so a redesign can be measured instead of eyeballed.
5. **Dedupe leak** (`CTX005`): `dedupe_native` inactive because the generated `CLAUDE.md`/`AGENTS.md` hash does not match (the EOL mismatch class from #247), reported with the fix command.
6. **Suggested shrinkers** (`--fix-plan`, no writes): for a layer over 90 %, list the top duplicate sentences and code fences by bytes that could be removed or moved to the vault. Never rewrite files automatically (the repo instructions make `.zirv/context/` the source of truth and `CLAUDE.md` generated; a linter that edits either fights that).

Wire `CTX001`/`CTX005` (errors) into `zirv context sync --report` and the workflow verify gate as a `CheckSpec` kind `context-lint`; the rest stay advisory.

## Files to touch

- new `src/commands/ctx/context_lint.rs` (pure; sentence splitter, normaliser, Jaccard, findings), `context.rs` (`lint` subcommand, report hook), `compile.rs` (expose the assembled layers with their budgets to the linter without re-reading), `src/commands/workflow/verification.rs` (`CheckKind::ContextLint`)
- fixtures: a pair of layers with a planted duplicate and a planted contradiction
- docs: `Modules/Built-in Commands.md`, `Modules/Ctx Subsystem.md` (context section)

## Acceptance criteria

- [ ] `zirv context lint` on this repo at the round-2 commit reports the known duplicate/contradiction cases from PR #261 (use the pre-fix texts as fixtures) and reports zero `CTX002`/`CTX003` after the fix.
- [ ] `common.md` at 4047/4096 bytes triggers `CTX001` warn; 4097 triggers error and the verify gate fails.
- [ ] Findings are deterministic and sorted (layer, byte offset); `--json` schema versioned.
- [ ] No file is ever written by `lint` or `--fix-plan`.
- [ ] Linter cost is bounded: sentence pairs capped (`context.lint_max_pairs`, default 20 000) with a `degraded` note when exceeded, matching the pre-computed budget-guard pattern.

## Tests

Pure tests per finding kind with fixtures; integration test over the repo's own `.zirv/context/`; clippy/fmt; full five gates before PR.

## Non-goals

- A quality grade or letter score (Ruflo's weights are unvalidated; ship measurements, not opinions).
- Auto-rewriting layers.

## Size / dependencies

S–M. Independent.

### LINKED ISSUE #293: Speed axis: per-turn wall-clock and time-to-first-token signals from transcript timestamps
## Origin

Prime Agent analysis, tracked in #279. `installAgentTelemetry` (`packages/coding-agent/src/core/telemetry.ts:604-794`) subscribes to session events and, per run, reports outcome, duration, `visible_ttft_ms` (time to the first non-empty `text_delta`, deliberately distinct from `first_model_event_ms`), summed and max model latency, and turn / tool / tool-error / compaction / retry counts, with provider and model reduced to bucketed categories. Verified wired (`agent-session-services.ts:262-268`, skipped for RLM children) and covered by 41 tests including "emits aggregate metrics without message or tool content" (`telemetry.test.ts:331`). The **metric set** is the takeaway. The transport is the anti-pattern: `getTelemetryEnabled` defaults `true` at all three layers (`settings-manager.ts:854-858`), so Prime ships an opt-out network sink to `api.primeintellect.ai`. zirv must not copy that — the split between "visible TTFT" and "first model event", and latency reported as a distribution rather than a mean, is what is worth having.

## Problem in zirv today

1. zirv has no wall-clock speed signal anywhere in the ctx event path. `NormalizedEvent` (`src/commands/ctx/event.rs:155-180`) carries `TurnStart`, `AssistantFinal { text, input_tokens }`, `ToolCall`, `ToolResult`, `ToolErrorText`, `Compaction` — not one timestamp, and `rot.rs` contains no `latency`/`elapsed`/`duration` reference at all.
2. The parser already exists and is unused for this: `window::parse_iso8601_utc` (`src/commands/ctx/window.rs:207`) is a bounded ISO-8601-to-unix-seconds parser with its own test battery (`window.rs:1384-1401`), used today only for usage windows.
3. `workflow/telemetry.rs` has a `duration_ms` field (`telemetry.rs:110`) populated per phase, but nothing per turn: `zirv workflow stats` (`telemetry.rs:882`) aggregates phase stats, cache-hit ratios and finding counts, so the intelligence and cost axes of the mandate are measurable and the speed axis is not.

### Timestamps as event fields, so `rot.rs` stays pure

Extend `NormalizedEvent::TurnStart` and `AssistantFinal` with `at: Option<u64>` (unix seconds, sub-second where the transcript gives it — store `at_ms: Option<u64>`), and add `at_ms` to `ToolCall`/`ToolResult`. Adapters fill these from the transcript line they already parse, via `parse_iso8601_utc` (extended with a milliseconds variant). `rot.rs` never calls a clock: time arrives as data on the event, exactly as `Capabilities::context_window_tokens` already delivers capacity. `ToolErrorText` stays fielded as-is.

New variant `AssistantFirstText { at_ms }`, emitted before `AssistantFinal` when the adapter can identify the first non-empty assistant text of the turn — the distinction Prime draws between visible TTFT and first model event. Adapters that cannot identify it emit nothing; absent is absent, never a guess.

### Derivation, in `score.rs`

`score.rs` (which already owns all I/O) derives per turn: `turn_ms` (TurnStart to AssistantFinal), `ttft_ms` (TurnStart to AssistantFirstText), tool-error rate (errored `ToolResult` over total). It reports p50 and max turn latency, p50 TTFT, and tool-error rate for the scored window. `rot.rs` gets nothing new to reason about in this issue — no new rot signal, no verdict change.

### Local recording

New `TelemetryKind::TurnLatencySampled` on `workflow/telemetry.rs`, with `turn_p50_ms`, `turn_max_ms`, `ttft_p50_ms`, `tool_error_rate` as `#[serde(default)] Option<...>` fields (schema stays additive; bump `TELEMETRY_SCHEMA_VERSION` only if a reader must reject old rows — it must not). `zirv workflow stats` gains a "speed" block. Retention, the 1000-event cap and the existing `REPO_FORBIDDEN` telemetry keys apply unchanged.

**Explicitly local-only.** No endpoint, no batching, no install UUID, no opt-out default. The state dir is the only sink, as it is for every other `workflow/telemetry.rs` event.

## Files to touch

- `src/commands/ctx/event.rs` (timestamp fields, `AssistantFirstText`)
- `src/commands/ctx/window.rs` (millisecond parse variant)
- `src/commands/ctx/adapters/claude.rs`, `adapters/codex.rs` (fill timestamps; codex may legitimately fill `None`)
- `src/commands/ctx/score.rs` (derivation, percentiles)
- `src/commands/workflow/telemetry.rs` (new kind and fields, stats block)
- `docs/obsidian/Modules/Rot Engine.md`, `docs/obsidian/Modules/Workflows.md`, `docs/obsidian/Modules/Ctx Adapters.md`, `docs/obsidian/Development/Work Journal.md`

## Acceptance criteria

- [ ] `rot.rs` still makes no fs/clock/env/net call; a `rot.rs` grep for `SystemTime`/`Instant` stays empty.
- [ ] Identical event streams (timestamps included) still give identical rot verdicts — adding timestamps changes no existing verdict.
- [ ] `score.rs` reports p50/max turn latency, p50 TTFT and tool-error rate for a fixture transcript with known timestamps.
- [ ] A transcript with unparseable or missing timestamps yields `None` metrics, never zero and never an error.
- [ ] TTFT is derived from first non-empty assistant text, and is absent (not equal to turn latency) when the adapter cannot identify it.
- [ ] `zirv workflow stats` prints the speed block; a state dir with only pre-change events prints it as "no data".
- [ ] No new network call: a repo-wide grep for a telemetry endpoint stays empty.

## Tests

Pure `event.rs`/`rot.rs` unit tests that the new fields do not perturb existing scoring. Fixture-transcript tests in `score.rs` for latency, TTFT, missing timestamps, out-of-order timestamps (clamp to zero, never negative) and a single-turn transcript. Adapter tests per harness for timestamp extraction. `telemetry.rs` round-trip tests including a legacy event file. No Linux Docker `wrap::` run required unless adapter argv changes, which it should not.

## Non-goals

- Any network telemetry, opt-out or otherwise.
- A rot signal or verdict driven by latency (that is a separate decision; #266 owns burn-slope).
- Benchmarking zirv's own startup.

## Size / dependencies

M. Feeds #266 (burn-slope rot signal — latency is a second series alongside token burn) and the sibling `zirv ctx measure` issue in this round (which reads the same derived metrics). Independent of #264.

### LINKED ISSUE #299: Prompt-prefix stability harness: assert the injected prefix is byte-stable across turns
## Origin

From Prime Agent analysis, tracked in #279. Prime's faux provider **emulates prompt caching**: it serializes the whole context, takes a common-prefix length against the previous prompt for the same `sessionId`, and reports real `cacheRead`/`cacheWrite` token numbers (`packages/ai/src/providers/faux.ts:192-226`, `commonPrefixLength` + `withUsageEstimate`). It is test-only by design and drives roughly 30 suite tests plus 55 regressions. The value is not the fake provider — it is that prefix stability becomes a *testable property* rather than a hope, and prefix stability is the single thing that actually moves cache cost. Prime gets it slightly wrong in one way zirv should not copy: its cache is keyed only on `sessionId` with no assertion that the prefix *should* be stable, so it measures drift without ever failing on it.

## Problem in zirv today

1. `tests/fixtures/fake-agent.sh` captures process behaviour only: argv (`FAKE_AGENT_ARGV_LOG`, line 68), `ZIRV_CTX_SESSION`/`WORK_GROUP`/`PARENT_SESSION`/`HEADLESS` env (lines 85-95), cwd (line 103), and a mode sequence. Nothing captures the assembled prompt, so no test anywhere asserts what the session actually received.
2. Nothing asserts the injected prefix is byte-identical from turn to turn. Every layer that can perturb it — memory harvest, the derived harness roster, mail, the ephemeral current-turn block — is free to move bytes into the prefix and silently void the harness's cache, and no gate would notice.
3. `.zirv/context/common.md` is 4047 bytes against `context.max_common_bytes = 4096` (`src/commands/ctx/config.rs:472`): 49 bytes of headroom, enforced only by the truncator at injection time, never by a test that fails the build.

**Prior art in zirv, already learned once.** `compile.rs`'s dedupe skip (`DEDUP_SKIP_ACTION`, `src/commands/ctx/compile.rs:63`; `native_file_already_carries_canonical`, `compile.rs:431`) requires a **byte-exact** match against the harness's native file, and its own test proves a CRLF-converted native file is not a match (`compile.rs:2302-2337`). The 3.3.0 EOL-pinning work (`.gitattributes`: inputs `eol=lf`, native files `-text`) exists purely to keep those bytes stable, and bought roughly a 25.9% prefix saving. That is exactly this issue's property, enforced for one file at one moment; this generalizes it to the whole prefix across turns.

### Fixture half

`tests/fixtures/fake-agent.sh` gains `FAKE_AGENT_PROMPT_LOG=<path>`: when set, each run appends the full assembled prompt it was handed (system prompt + `common.md` + harness layer + roster + context + memory + any per-turn block), byte-exact, framed as `\x1e<turn-index>\x1e` followed by the raw bytes. No normalization, no trailing newline added, no shell interpolation of the payload — the log must be a faithful byte record or the test it feeds is worthless. The fixture is POSIX `sh` and must keep working under the Linux Docker PTY suite.

### Assertion half

Inline tests in `src/commands/ctx/compile.rs`'s and `prompt.rs`'s `#[cfg(test)] mod tests`:

1. **Prefix identity.** Across N turns of one session, the bytes up to the declared prefix boundary are identical. Boundary = everything `compile_with_harness_roster` emits before the ephemeral current-turn block.
2. **Declared suffix.** A table of state changes — memory harvest, roster refresh, mail arrival — each asserted to perturb *only* the suffix. A change that moves the boundary fails with the layer name, so adding a layer forces an explicit decision about which side it belongs on.
3. **Budget.** `include_str!(".zirv/context/common.md").len() < 4096`, with the current headroom printed on failure.

### Failure message

Failure names the **first differing byte offset**, plus 32 bytes either side of it from both prompts rendered as escaped bytes, plus which layer owns that offset (from `CompiledContext`'s existing provenance). "Prefix drifted" with no offset is not actionable.

### Purity

Pure assertions over already-captured text. `rot.rs` is untouched; no runtime behaviour changes; nothing is added to any shipped code path.

## Files to touch

- `tests/fixtures/fake-agent.sh`
- `src/commands/ctx/compile.rs`, `src/commands/ctx/prompt.rs` (test modules; a small shared comparison helper)
- `docs/obsidian/Concepts/Context Management.md`, `docs/obsidian/Development/Testing Guide.md`

## Acceptance criteria

- [ ] `FAKE_AGENT_PROMPT_LOG` records one byte-exact framed record per turn.
- [ ] A test fails if the prefix differs between turn 1 and turn N of one session.
- [ ] A memory harvest between turns perturbs the suffix only; a test proves it.
- [ ] A roster refresh between turns perturbs the suffix only; a test proves it.
- [ ] A deliberately injected one-byte prefix change fails with the correct first-differing offset and owning layer.
- [ ] `common.md` exceeding 4096 bytes fails a test, not just the truncator.

## Tests

The issue *is* tests. Inline pure tests for the comparison/offset helper; fixture-driven tests for the multi-turn logs. The `fake-agent.sh` change must be verified on Linux/Docker before pushing: export with `git -c core.autocrlf=false archive HEAD` and run `cargo test --bin zirv wrap:: -- --test-threads=1` on `rust:1-bookworm` as a non-root user (plain `git archive` emits CRLF and corrupts shell fixtures).

## Non-goals

- Comparing *outcomes* between prompt variants — that is #277 (prompt-layer A/B harness). This is a byte-level invariant with no model in the loop.
- A scripted per-turn response queue like Prime's faux provider.
- Measuring real provider cache hits.

## Size / dependencies

M. Depends on nothing; pairs with issue 19 in this batch (its per-session probe cache exists precisely so a mid-session install cannot flap the roster line and break this invariant). Distinct from #275 and #277.

### LINKED ISSUE #306: Safety hook prompts on routine local git actions: path checkout/restore, scoped clean, agent-owned worktree remove, rebase
## Problem

On an interactive launch the safety hook is the sole permission gate (`--permission-mode default`), so every built-in `Ask` verdict becomes a prompt to the operator. In one orchestration session (2026-09-02/03, zirv 3.11.0) the operator was prompted four times, every time by the built-in `<vcs: destructive local or remote action>` classifier in `src/commands/ctx/safety.rs` (`is_destructive_vcs_action`) on routine, local, agent-scoped git work:

| # | command (abridged) | rule | what it actually was |
|---|---|---|---|
| 1 | `git worktree remove <scratchpad>/main-wt --force && ...; git worktree list` | worktree remove `--force` | removing a scratch worktree the agent itself created under the scratchpad root |
| 2 | `git clean -fdq .zirv/work/<uuid>` | clean with `-f` | deleting one workflow artifact directory zirv itself created (`rm -rf` on it is a built-in Deny, so this was the only remaining way) |
| 3 | `git merge --no-ff --no-commit <branch>; git checkout HEAD -- Cargo.toml Cargo.lock && git commit ...` | `checkout ... -- <paths>` | dropping a worker's version bump from a merge: two named tracked files |
| 4 | same shape as 3 for the next track | same | same |

The hook's own retry boundary already treats a tracked-file checkout as acceptable next to allowed git inspection (`retry_has_allow_verdict` / `is_checkout_path_restore`), so the main evaluation path is stricter than the retry path for the same command.

## Proposal (built-in defaults; no operator or repo config change)

Narrow `is_destructive_vcs_action` so that local, recoverable or agent-scoped actions are not "destructive":

- `checkout [<rev>] -- <paths>` and `restore [<flags>] <paths>`: Allow when every path is a concrete path (not `.`, `:/`, `*`, or a glob); keep Ask for tree-wide forms.
- `clean -f...` with one or more explicit path arguments and without `-x`/`-X`: Allow; keep Ask for a bare `git clean -f`/`-fdx` (no path, or ignored files).
- `worktree remove --force <path>` whose path lies under an agent-owned root: `.claude/worktrees/`, `.zirv/worktrees/`, or a scratchpad root (`target_is_confined`): Allow; keep Ask elsewhere.
- `rebase` (non-interactive): Allow; it is local and reflog-recoverable. `filter-branch` stays Ask.

Everything else in the classifier is unchanged: force/delete pushes, `reset --hard`, `branch -D`, `stash drop|clear`, `reflog expire|delete`, `gc --prune=now`.

## Acceptance

- The four commands above evaluate to Allow interactively and headlessly (table test).
- `git clean -fdx`, `git clean -fd` (no path), `git checkout -- .`, `git restore .`, `git worktree remove ../feature --force`, `git push --force`, `git reset --hard` still Ask.
- `docs/obsidian/Modules/Command Safety.md` lists the narrowed forms.

Glubiz:
## Release 3.12.0: second harness cost/speed/intelligence batch

Closes #267
Closes #275
Closes #264
Closes #293
Closes #299
Closes #306

Selected from the twelve remaining open cost/speed/intelligence issues after verifying each premise against `main` at 985b5e5; implemented in six worktree tracks and merged here. Version 3.11.0 -> 3.12.0.

### What changed

- **#267 Reader and writer permits (speed, intelligence).** `zirv ctx agent --mode read-only|writing` (default `writing`) and `--worktree`. A second permit pool `supervise.max_writers` (default 1, `REPO_FORBIDDEN`) reuses the heavy-permit file scheme and dead-owner sweep; a writing worker holds a writer permit keyed by canonical tree path for its lifetime, so a second writer in the same checkout is refused with `EXIT_WRITER_BUSY` and a one-line reason while read-only workers share any tree. `--worktree` allocates `<repo>/.zirv/worktrees/<short>` from the session base. `Delegation.mode` and `SpawnRequest.mode` are additive; review/test/verify auto-spawns are read-only; the dashboard spawn path enforces the same pool; both orchestrator prompts gain one integration-owner sentence.
- **#275 Context lint (cost, intelligence).** `zirv context lint [--json] [--budget] [--fix-plan]` reports CTX001 budget headroom, CTX002 duplicate imperative sentences across layers, CTX003 contradiction candidates, CTX004 proportionality, CTX005 dedupe leaks, deterministically and without writing anything. CTX001/CTX005 errors fail `context sync --report` and a new `context-lint` verify check discovered whenever `.zirv/context/` exists. `context.lint_max_pairs` (default 20 000, `REPO_FORBIDDEN`) bounds the pair scan. `CompiledContext::emitted_layers()` exposes layers with byte ranges and budgets.
- **#264 Cost ledger (cost).** Pure `price.rs` (built-in table `as_of 2026-09-01`, integer micro-USD, `~/.zirv/prices.toml` override, `price.stale_after_days`/`price.table_path` operator-only). `zirv ctx spend [--session|--group|--since] [--by harness|model|task-class|worker] [--json]` aggregates `delegations.jsonl` deterministically. `Delegation.task_class` (`--task-class`, derived for auto-spawns), telemetry `cost_micros`/`price_as_of` and a populated `parent_session_id`, a cost block in `zirv workflow stats`, one `spend:` line in `zirv ctx status`, and a dashboard aggregate row under a source contract that renders `--` without a live source.
- **#293 Speed signals (speed).** `NormalizedEvent` carries `at_ms` on `TurnStart`/`AssistantFinal`/`ToolCall` plus `AssistantFirstText` and `ToolResultTimestamp` variants; both adapters fill them (`window::parse_iso8601_utc_ms`). `score::derive_speed_metrics` reports p50/max turn latency, p50 TTFT and tool-error rate; missing timestamps give `None`. `TelemetryKind::TurnLatencySampled` is recorded once per Stop-hook turn and `zirv workflow stats` prints a speed block. `rot.rs` stays pure and a pinned test proves identical verdicts with and without timestamps.
- **#299 Prompt-prefix stability harness (cost).** `FAKE_AGENT_PROMPT_LOG` makes the fake agent append a byte-exact framed record of the prompt it was handed per turn. Tests assert the composed prefix is byte-identical across compiles, that memory harvest, roster refresh and mail arrival perturb only the suffix, that a one-byte prefix change fails with the exact offset and owning layer, and that `common.md` stays under its 4096-byte budget (headroom printed on failure). No shipped code path changes.
- **#306 Safety hook prompts on routine git (filed from this session's retrospective).** The built-in `<vcs: destructive local or remote action>` classifier prompted four times in one session on `git checkout HEAD -- <files>` during merges, a scoped `git clean -fdq .zirv/work/<id>`, and `git worktree remove --force` of a scratch worktree. Narrowed: concrete-path `checkout -- <paths>`/`restore <paths>`, `clean -f <paths>` without `-x`, `worktree remove --force` under `.claude/worktrees/`, `.zirv/worktrees/` or a scratchpad root, and non-interactive `rebase` are no longer destructive; tree-wide forms, `-x`, arbitrary `--force` targets, force/delete pushes, `reset --hard`, `branch -D`, `filter-branch` still ask. The blanket `Bash(git rebase *)`/`Bash(git clean *)` entries left `SHIPPED_POSTURE_ASK` so the per-argument judgment is reachable.

### Not taken this round

#294 (`zirv ctx measure`, L, consumes #293's metrics), #295 (session-tier default for `remember` is an operator product decision), #262 (authority narrowing, a safety posture rather than one of the three axes), #272/#276/#302 (hygiene), #260 (Windows-only test bug), #303 (rejected with evidence in 3.11.0).

### Verification

- Windows (final head c5b5725): `cargo build`, `cargo nextest run --no-fail-fast`, `cargo test --verbose -- --test-threads=1`, `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings`. The serial run fails only on this machine's pre-existing set, each confirmed failing on `main`: `adapters::claude::launch_settings_are_materialized_atomically_under_the_zirv_home`, `adapters::tests::harness_prompt_lines_never_borrows_a_named_adapters_presence_for_another`, `adapters::tests::liveness_probe_is_unknown_when_a_path_entry_is_not_a_directory`, `wrap::win::a_supervised_wrap_binds_a_turn_signal_transport`, `supervise::terminate_{pid_}stops_*`, plus the load-flaky `supervise::a_deadline_kills_the_child_and_reports_a_timeout` (a Docker build ran alongside that serial phase; it passes in isolation and passed serially on the two previous heads). fmt and clippy clean.
- Linux (`rust:1-bookworm`, non-root, autocrlf=false archive) on c5b5725: full serial suite 3922 passed / 0 failed, clippy clean. Covers the `#[cfg(unix)]` fixture tests for `fake-agent.sh`, the dash lineage tests and the adapter timestamp parsing.
- Workflow 6e07db72: intent/spec/plan approved, test gate passed on the baselined names only. Review via `zirv workflow review run` on claude (sonnet) and codex, three rounds: round 1 found 4 findings (`git clean -fd .` allowed by the narrowed classifier; `--worktree` never reclaimed; per-tree writer claim not atomic with `max_writers > 1`; invalid `--since` silently unfiltered), round 2 found 3 (`..`/`./` counted as concrete clean paths; tree claim did not follow `set_child_pid`; worktree leaked on dashboard handoff and pre-launch failures), round 3 found 2 (reclaim guard armed on the unconfirmed-dashboard exit; worktree ownership inferred from the path prefix). All 9 fixed and disposed; the round-3 fixes (c5b5725) were verified by the gates above, not by a fourth reviewer pass, because the workflow's review loop is capped at three rounds. Dashboard coordinator panes no longer take a writer permit (caught by the Linux run).
- Vault pages updated for every behaviour/contract change; `.zirv/worktrees/` added to `.gitignore`.

https://claude.ai/code/session_01KTtYgG7JTr6EGuFrpvVCVg


Glubiz:
Superseded by #324: identical tree (c5b5725) replayed as a linear history so the branch can be rebase-merged; the merge commits here blocked GitHub's rebase.

PR #341, opened 2026-09-04
feat(memory): operator-owned global memory scope (#336)

### LINKED ISSUE #336: Global memory bank: operator-owned facts shared by every session on the machine, not pinned to a repo
## Problem

zirv has two memory banks and both are pinned to one repository:

| Bank | Location | Scope key | Trust |
|---|---|---|---|
| Private | `<state>/memory/<repo_slug>/` (`~/Library/Application Support/zirv/ctx/memory/...` on macOS) | `state::repo_slug(repo)` | operator-owned, trusted |
| Shared | `<repo>/.zirv/memory/` | the checkout itself | repo-owned, untrusted |

`StateDir::memory()` returns `<state>/memory`, and every read/write path in `memory.rs` immediately joins `repo_slug` onto it (`MemoryScope::dir`, `upsert`, `list`, `forget`). There is no bank for facts that are true for every session on this machine regardless of which folder it was launched from.

Observed on this machine (2026-09-04): the private bank holds 14 entries under the zirv-dynamic-cli slug and 1 under the CRM slug. Facts like "codex-cli is installed at `<path>`", "GitHub CI runners have no harness binaries, stub them", "never `taskkill` a `zirv*.exe`, this session runs under zirv", "PATH must be `$HOME/.cargo/bin:/usr/bin:/bin` when re-running tests locally", and operator preferences ("dispatch `zirv agent` as visible dashboard panes, never `--headless`", "no `Co-Authored-By` lines") are machine- or operator-level. Today they either get re-learned per repo, get duplicated into each repo's private slug by hand, or live in Claude Code's own `~/.claude/projects/<slug>/memory/` which zirv does not manage and Codex sessions never see.

## Proposal

Add a third, operator-owned scope, `MemoryScope::Global`, stored at `<state>/memory/_global/` (or a sibling `<state>/global-memory/`; pick whichever keeps `repo_slug` collisions impossible, since a slug is derived from a path and could in principle be `_global`).

- Same `Entry` markdown format and parser as the two existing scopes; only the storage root differs.
- Trust level: operator-owned, same as `Private`. It is never read from a repository checkout, so none of the shared-scope hardening (symlink refusal, credential-key refusal, key-conflict suppression against private) is needed against it. It DOES need the same key-conflict rule *in the other direction*: a repo-owned shared entry whose key matches a global entry is dropped, exactly as shared-vs-private is today.
- Precedence in `prompt::select_memory_within_cap`: private (repo) > global > shared. A repo-specific fact should beat a machine-wide one on key conflict; both beat untrusted repo content.
- Injection: `with_memory_layer` renders global entries alongside private ones under the existing cap. Consider a separate small sub-cap so a large global bank cannot starve repo-specific private entries.
- CLI: `zirv ctx remember|recall|forget --global` and `zirv memory remember|list|recall|forget --global`, mirroring the existing `--repo` / `--shared` selectors via `MemoryScope::of`-style mapping (that helper currently maps one bool; it will need to become a two-flag or enum selector with `--repo` and `--global` mutually exclusive).
- `zirv memory status` reports the global bank's path, entry count, and enabled state like the other two.
- Config: one new operator-only key, `memory.global_enabled` (default `true`, underneath the `memory.enabled` master switch, `REPO_FORBIDDEN` with a `ZIRV_CTX_MEMORY_GLOBAL` env override) so an operator can turn it off without losing repo memory.
- Harvest: automatic harvest keeps writing to the shared/private banks only. Global entries are explicit (`Source: explicit`) until there is a signal that a harvested fact is machine-wide, which there is not today.
- `zirv ctx optimize` / `memory_optimize` should include the global bank in low-value-entry findings and offer a "promote to global" suggestion when the same key+body appears in two or more repo slugs' private banks on the same machine.

## Acceptance criteria

- [ ] `zirv ctx remember --global k "v"` writes `<state>/memory/_global/<ts>-k.md` (or the chosen root) and never touches any `<repo_slug>` directory.
- [ ] A session launched from repo A and one from repo B both receive the global entry in their memory layer; neither receives the other's private entries.
- [ ] Key conflict: private beats global, global beats shared, all case-insensitive; tests shaped like `a_shared_entry_reusing_a_private_keys_key_is_dropped_not_merely_outranked` and its different-case sibling in `prompt.rs`.
- [ ] `memory.enabled = false` disables all three banks; `memory.global_enabled = false` disables only the global one; both keys hard-error in the repo layer.
- [ ] `zirv memory status` lists the global bank; `zirv ctx recall --json` tags entries `"scope": "global"`.
- [ ] `--repo` and `--global` together is a clap conflict error.
- [ ] Existing private/shared tests unchanged; five gates pass.

## Notes

- The Codex adapter should receive the same layer; today the memory layer is provider-neutral in `prompt.rs`, so this should fall out for free, but assert it.
- Windows: the state dir resolves via `dirs::data_local_dir`; keep the same root there. The verbatim-prefix stripping fixed in #333 does not apply since no repo path is involved.

---
Environment
- Zirv: 3.15.1
- OS: macos
- Architecture: aarch64
- Harness: claude
- Model: fable

Glubiz:
Closes #336.

Adds a third memory scope, `MemoryScope::Global`: an operator-owned, machine-wide bank at `<state>/memory/_global/` that every zirv-supervised session receives regardless of which repository it was launched from.

- Trusted exactly like the private bank; never read from a repository checkout. Gated only by the master `memory.enabled` switch (operator ruling: no global-only toggle).
- Precedence private > global > shared in one byte cap. Key conflicts are suppressed case-insensitively before ranking: a global key that matches a private key is dropped, a shared key that matches a private or global key is dropped. The shared block's header, screening and end marker are unchanged.
- `prompt::MemoryLine.shared: bool` becomes `scope: MemoryScope`; global lines render inside the trusted block after private lines.
- CLI: `zirv ctx remember|forget --global` (conflicts with `--repo`), `zirv memory list|recall|remember|forget|verify --global` (conflicts with `--shared`); `zirv ctx recall` lists all three scopes with `"scope": "global"` in JSON; `zirv ctx forget --all --global` clears only the global bank; `zirv memory status` reports the global bank.
- `GLOBAL_SLUG = "_global"` cannot collide with `state::repo_slug` output (tested).
- Out of scope, unchanged: harvest never writes to the global bank; `zirv memory optimize` ignores it.

Version 3.15.1 -> 3.16.0.

Verification (macOS, sandboxed): build, fmt, clippy clean; `cargo nextest run --no-fail-fast` name-diffed against a clean 0ddfc58 worktree in the same environment: no new failures (the only differing name is the known exec timing flake, which passed on two immediate reruns); serial `cargo test -- --test-threads=1` name-diffed the same way. Independent sonnet review: no findings.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2693,7 +2693,7 @@ checksum = "b97154e67e32c85465826e8bcc1c59429aaaf107c1e4a9e53c8d8ccd5eff88d0"
 
 [[package]]
 name = "zirv"
-version = "3.15.1"
+version = "3.16.0"
 dependencies = [
  "clap",
  "console",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zirv"
-version = "3.15.1"
+version = "3.16.0"
 edition = "2024"
 authors = ["Jonathan Solskov <josj@zirv.io>"]
 description = "A CLI tool for executing developer-defined YAML scripts with support for interactive commands and OS-specific execution."
--- a/src/commands/ctx/compile.rs
+++ b/src/commands/ctx/compile.rs
@@ -367,16 +367,21 @@ fn gather_memory(
 ) -> (Vec<prompt::MemoryLine>, Vec<prompt::MemoryLine>) {
     // Read every memory-bank `.md` file once and hand the same in-memory
     // entries to both consumers below -- `render_for_prompt`/`candidates_
-    // for_repo` each scan the identical private+shared bank on their own,
+    // for_repo` each scan the identical private+global+shared bank on their own,
     // which used to mean every file was read twice on every session launch
     // (see `memory::LoadedMemory`'s own doc comment).
-    let loaded = memory::load_both_scopes(repo, state, slug, cfg);
+    let loaded = memory::load_all_scopes(repo, state, slug, cfg);
     let core = memory::render_for_prompt_from_loaded(&loaded);
     let core_keys: std::collections::HashSet<(bool, String)> =
         prompt::select_memory_within_cap(&core, cfg.memory.core_max_bytes)
             .0
             .into_iter()
-            .map(|entry| (entry.shared, entry.key.to_lowercase()))
+            .map(|entry| {
+                (
+                    entry.scope == memory::MemoryScope::Shared,
+                    entry.key.to_lowercase(),
+                )
+            })
             .collect();
 
     let candidates = retrieval::candidates_from_loaded(&loaded, now);
@@ -413,7 +418,11 @@ fn gather_memory(
             body: ranked.candidate.entry.body.clone(),
             verified: ranked.candidate.entry.verified,
             written: ranked.candidate.entry.written,
-            shared: ranked.candidate.shared,
+            scope: if ranked.candidate.shared {
+                memory::MemoryScope::Shared
+            } else {
+                memory::MemoryScope::Private
+            },
         })
         .collect();
     (core, retrieved)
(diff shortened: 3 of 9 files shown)


Glubiz:
Superseded by the single release PR #343 (release/3.17.0), which carries these exact commits on top of main 8a7710c. Kept as a draft for reference; close once #343 lands.

